PR #13, opened 2019-03-01
Improves overall handling of templates

### LINKED ISSUE #11: Improve handling of C++ template functions
The `rellic-headergen` tool should produce a functioning C header equivalent from the following code.
```C++
#include <utility>

class MyClass {
  std::pair<int, int> my_pair;
  std::pair<int, int> MyMethod(std::pair<int, int> pair);
};
```
So far it produces
```C++
struct pair {
    int first;
    int second;
};
void _ZNSt4pairIiiEC1Ev(struct pair *this);
void _ZNSt4pairIiiEC1ERKiS2_(struct pair *this, const int &__a, const int &__b);
struct MyClass {
    std::pair<int, int> my_pair;
};
std::pair<int, int> _ZN7MyClass8MyMethodESt4pairIiiE(struct MyClass *this, std::pair<int, int> pair);
```
A good way to evaluate the results is to compare the LLVM IR produced by code that uses the original C++ header and the generated C header. The shape and size of `class.MyClass` in the LLVM IR from both headers should be the same. Name mangling should be the same as well.

surovic:
Resolves #11

### DIFF
--- a/rellic/AST/CXXToCDecl.cpp
+++ b/rellic/AST/CXXToCDecl.cpp
@@ -17,7 +17,7 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 
-#include <clang/Index/CodegenNameGenerator.h>
+#include <clang/AST/Mangle.h>
 
 #include "rellic/AST/CXXToCDecl.h"
 #include "rellic/AST/Util.h"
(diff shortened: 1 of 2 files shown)


surovic:
Currently we're able to go from this
```c++
#include <utility>

class MyClass {
public:
  std::pair<int, int> my_pair;
  std::pair<int, int> MyMethod(std::pair<int, int> pair);
};
```
to this
```c
struct _ZTSSt4pairIiiE {
    int first;
    int second;
};
void _ZNSt4pairIiiEC1Ev(struct _ZTSSt4pairIiiE *this);
void _ZNSt4pairIiiEC1ERKiS2_(struct _ZTSSt4pairIiiE *this, const int &__a, const int &__b);
struct _ZTS7MyClass {
    struct _ZTSSt4pairIiiE my_pair;
};
struct _ZTSSt4pairIiiE _ZN7MyClass8MyMethodESt4pairIiiE(struct _ZTS7MyClass *this, struct _ZTSSt4pairIiiE pair);
```

PR #25, opened 2019-04-10
Fixes an assert / infinite loop / sigsegv in CreateIfStmt()

surovic:


surovic:
There's a change at [line 29](
https://github.com/trailofbits/rellic/pull/25/commits/8f48996527254317d8d50e69aa3c7db6cd0f8fe2#diff-fd36dc2fb4e941cf3bfc45379447c5a4R29)
```C
ifstmt->setElse(nullptr);
```
In clang-8.0 the `Create` static method for `clang::IfStmt` was introduced and it changes how memory is allocated for the optional part of a condition. Trying to call `clang::IfStmt::setElse()` with a condition object that was not created with memory for the `else` branch allocated results in an assert in debug builds and undefined behavior in releases.

PR #33, opened 2019-07-05
Translation and simplification of aggregate LLVM types

### LINKED ISSUE #29: Handling structured LLVM types
Assertion reproducible by [sqlite.zip](sqlite.zip).

surovic:
Resolves #29

Adds initial handling of aggregate LLVM types and NULL pointer constants. This includes translation of GEP instructions with aggregate type operands to the corresponding C operations (array subscript and member access expressions).

Subsequently adds Z3 simplification and rule-based simplification of expressions with above mentioned operands.

### DIFF
--- a/rellic/AST/ExprCombine.cpp
+++ b/rellic/AST/ExprCombine.cpp
@@ -82,9 +82,9 @@ class AddrOfArraySubscriptRule : public InferenceRule {
 class DerefAddrOfRule : public InferenceRule {
  public:
   DerefAddrOfRule()
-      : InferenceRule(unaryOperator(stmt().bind("deref"), hasOperatorName("*"),
-                                    has(unaryOperator(hasOperatorName("&"))))) {
-  }
+      : InferenceRule(unaryOperator(
+            stmt().bind("deref"), hasOperatorName("*"),
+            has(ignoringParenCasts(unaryOperator(hasOperatorName("&")))))) {}
 
   void run(const MatchFinder::MatchResult &result) {
     match = result.Nodes.getNodeAs<clang::UnaryOperator>("deref");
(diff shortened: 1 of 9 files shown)


rellic/AST/ExprCombine.cpp
```diff
@@ -82,9 +82,9 @@ class AddrOfArraySubscriptRule : public InferenceRule {
 class DerefAddrOfRule : public InferenceRule {
  public:
   DerefAddrOfRule()
-      : InferenceRule(unaryOperator(stmt().bind("deref"), hasOperatorName("*"),
-                                    has(unaryOperator(hasOperatorName("&"))))) {
-  }
+      : InferenceRule(unaryOperator(
+            stmt().bind("deref"), hasOperatorName("*"),
+            has(ignoringParenCasts(unaryOperator(hasOperatorName("&")))))) {}
```
pgoodman: Can it happen that `ignoringParenCasts` will accept an outermost `(uintptr_t)`, or does that imply something structurally unacceptable? That is, we're in a deref, therefore we must be dereferencing an address-of.

rellic/AST/ExprCombine.cpp
```diff
@@ -95,12 +95,13 @@ class DerefAddrOfRule : public InferenceRule {
     auto deref = clang::cast<clang::UnaryOperator>(stmt);
     CHECK(deref == match)
         << "Substituted UnaryOperator is not the matched UnaryOperator!";
-    auto addr_of = clang::cast<clang::UnaryOperator>(deref->getSubExpr());
+    auto subexpr = deref->getSubExpr()->IgnoreParens();
```
pgoodman: Can types be an issue here? E.g. `*(uint8_t *)&int_val;`?

pgoodman: Do you think it'd be reasonable, where the types don't match, to substitute with the appropriate C++ case type, i.e. `static_cast`, `reinterpret_cast`, or even fall back to a `memcpy` in some cases, and if the desired output is C, then those more explicit casting forms can be changed later?

rellic/AST/ExprCombine.cpp
surovic: As far as I know, any C-style cast will disqualify the match, since `ignoringParenCasts` does not ignore C-style cast expressions.

> That is, we're in a deref, therefore we must be dereferencing an address-of.

This is correct. An address-of that can be wrapped in cast expressions that `ignoringParenCasts` catches.

rellic/AST/ExprCombine.cpp
surovic: Type casts will be an issue here in the sense that the matcher of this rule will not match `*(uint8_t *)&int_val` since there's an extra C-style cast expression that is not ignored.

PR #42, opened 2019-11-01
Adds tests to CMake and Travis CI

### LINKED ISSUE #2: Set up Travis CI with tests
There are some basic in the repository's `tests` folder. Automating the test using something like [pytest](https://docs.pytest.org/en/latest/) would be great.

surovic:
resolves #2


scripts/roundtrip.py
```diff
@@ -41,7 +41,7 @@ def compile(clang, input, output, timeout, options=None):
     cmd.extend([input, "-o", output])
     p = run_cmd(cmd, timeout)
     assert p.returncode == 0, \
```
pgoodman: Only use `assert` for things that you optionally want to check, just in case this is run with "optimizations" enabled. I.e. turn these `assert`s into actual `if` conditions that raise exceptions.

surovic: Changed `assert` to `unittest.TestCase.assertEqual()`. This should hopefully survive optimized runs.

PR #45, opened 2019-11-11
Add more missing BinOp

meme:
- `Or` and `Mul` opcodes
- Added print so the missing opcode is known in the error

Note that division is still missing, I am not certain on how to handle both unsigned or signed division (or does it even matter, I may just lift it as the `clang::BO_Div`?)

### DIFF
--- a/rellic/AST/IRToASTVisitor.cpp
+++ b/rellic/AST/IRToASTVisitor.cpp
@@ -670,6 +670,10 @@ void IRToASTVisitor::visitBinaryOperator(llvm::BinaryOperator &inst) {
       binop = BinOpExpr(clang::BO_Rem, lhs->getType());
       break;
 
+    case llvm::BinaryOperator::SRem:
+      binop = BinOpExpr(clang::BO_Rem, lhs->getType());
+      break;
+
     case llvm::BinaryOperator::Add:
       binop = BinOpExpr(clang::BO_Add, type);
       break;
@@ -683,7 +687,7 @@ void IRToASTVisitor::visitBinaryOperator(llvm::BinaryOperator &inst) {
       break;
 
     default:
-      LOG(FATAL) << "Unknown BinaryOperator: " << llvm::Instruction::getOpcodeName(inst.getOpcode());
+      LOG(FATAL) << "Unknown BinaryOperator: " << inst.getOpcodeName();
       break;
   }
 }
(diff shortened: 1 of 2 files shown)


rellic/AST/IRToASTVisitor.cpp
```diff
@@ -674,8 +678,12 @@ void IRToASTVisitor::visitBinaryOperator(llvm::BinaryOperator &inst) {
       binop = BinOpExpr(clang::BO_Sub, type);
       break;
 
+    case llvm::BinaryOperator::Mul:
+      binop = BinOpExpr(clang::BO_Mul, type);
+      break;
+
     default:
-      LOG(FATAL) << "Unknown BinaryOperator operation";
+      LOG(FATAL) << "Unknown BinaryOperator: " << llvm::Instruction::getOpcodeName(inst.getOpcode());
```
surovic: Probably better to use `inst.getOpcodeName()` directly here.

meme:
Ideally I just go for it and add every missing operator?

surovic:
That would be great. But ideally we would also want to have some basic test to go with it. You can use stuff in `tests/tools/decomp/` as inspiration.

meme:
Just a question about the test suite, how do I run it, and how does it verify the decompilation is correct? Or is it just like a collection of code samples?

surovic:
> ...how does it verify the decompilation is correct?... 

So what we do is essentially a roundtrip-type test. We take C code, compile it first to an executable, run it and record the outputs (stdout, stderr, return code). Then we take the same C code, compile it down to bytecode, decompile the bytecode with rellic back to C. Then we compile the decompiled C code down to an executable, run it and record outputs. The outputs of both executables should be the same. The test generator and runner is implemented in [scripts/roundtrip.py](https://github.com/lifting-bits/rellic/blob/master/scripts/roundtrip.py).

As for how to run the whole thing, there's a `make test` target in our CMake that'll do the trick.

pgoodman:
@surovic Merging is up to you.

meme:
Just added the remaining binary operators as per: https://github.com/llvm-mirror/llvm/blob/master/include/llvm/IR/Instruction.def#L147 (I have omitted the floating point operations), I'll push shortly and add a division test go along with it.

meme:
Apologies, but can we revert this? I'm not done just yet...

meme:
I also realized that I did not add these operators to the Z3 visitor, so this will result in further problems if the operators are used as predicates. I will address this, along with other comparison operators, because why not.

surovic:
I think it's gonna be cleaner if the rest of the binops is in a new PR. What do you think?

meme:
https://github.com/lifting-bits/rellic/pull/52

PR #48, opened 2019-11-20
Add Docker image and documentation

ekilmer:
There is no testing set up for the Docker image, and your Travis scripts look pretty sophisticated, so I decided not to touch that. If you could describe how you think tests should work, I can work with you to implement that!

Let me know if you think there is anything missing!


ekilmer:
Ping @artemdinaburg to test if you need anything else!

README.md
```diff
@@ -74,3 +74,35 @@ To try out Rellic you can do the following, given a LLVM bitcode file of your ch
...
+docker build -t rellic-decomp-40 --build-arg LLVM_VERSION=4.0 .
+```
+
+To run the decompiler, the entrypoint has already been set, but make sure the bitcode you are decompiling is the same as LLVM version as the decompiler, and run:
+
+```sh
+docker run --rm -t -i \
+  -v $(pwd):/test -w /test \
+  -u $(id -u):$(id -g) \
+  rellic-decomp-40 --input ./tests/tools/decomp/issue_4.bc --output /dev/stdout
```
surovic: The `tests/tools/decomp/` folder contains C sources. The compilation of `issue_4.bc` is missing here.

surovic: Maybe also move the `docker-decomp-entry-point.sh` to `scripts/` if possible. Just to keep things neat. Unless there's some convention of such scripts being in the root of the repo. I really don't have too much experience with setting up docker images.

surovic:
>If you could describe how you think tests should work, I can work with you to implement that!

By tests, you mean rellic tests? If so, there's a target in the CMake generated makefiles. Running `make tests` in the build folder should do the trick.

README.md
ekilmer: I added a command to build the bc file in the README.

Also moved `docker-decomp-entry-point.sh` to `scripts/`

ekilmer:
> By tests, you mean rellic tests? 

Yes! I think I figured that out now. Had to add `python3.7` dependency, but looks like it should work now.

I aso want to add something to Travis to test the Dockerfile and make sure it doesn't fall out of date.

surovic:
Maybe we could just run the `travis.sh` script in a ubuntu 16.04 docker image that would be hosted in the travis VM? It's kind of a dirty solution, but I guess it would work?

pgoodman:
@surovic Were the requested changes resolved to your satisfaction?

ekilmer:
Finally, I think I got travis test to work with the Docker image test

surovic:
@ekilmer Looks good to me! Any other stuff you want to add, or should we merge?

ekilmer:
@surovic This should be good to go!

I also noticed that the Docker image is a bit heavy at around 2GB, since I just copy the build directory into the final image. Most of the data is likely due to all the dependencies downloaded during the build process, and I don't see an `install` target to get the minimal set of binaries, libraries, headers, etc.. We will want to change the Dockerfile if an `install` target is added in the future.

surovic:
@ekilmer can you make an issue for the install target and this rationale? Just so it's recorded and not forgotten.

PR #52, opened 2019-11-22
Add more comparison operators (and add to Z3)

meme:
- Add comparison operators to Z3
- Add binary operators to Z3
- Add tests for binary operators

Continuation of https://github.com/lifting-bits/rellic/pull/45 as per https://github.com/lifting-bits/rellic/pull/45#issuecomment-557729374

### DIFF
--- a/rellic/AST/Z3ConvVisitor.cpp
+++ b/rellic/AST/Z3ConvVisitor.cpp
@@ -532,12 +532,20 @@ bool Z3ConvVisitor::VisitBinaryOperator(clang::BinaryOperator *c_op) {
   auto rhs = GetOrCreateZ3Expr(c_op->getRHS());
   // Create z3 binary op
   switch (c_op->getOpcode()) {
-    case clang::BO_LAnd:
-      InsertZ3Expr(c_op, Z3BoolCast(lhs) && Z3BoolCast(rhs));
+    case clang::BO_GT:
+      InsertZ3Expr(c_op, lhs > rhs);
       break;
 
-    case clang::BO_LOr:
-      InsertZ3Expr(c_op, Z3BoolCast(lhs) || Z3BoolCast(rhs));
+    case clang::BO_LT:
+      InsertZ3Expr(c_op, lhs < rhs);
+      break;
+
+    case clang::BO_GE:
+      InsertZ3Expr(c_op, lhs >= rhs);
+      break;
+
+    case clang::BO_LE:
+      InsertZ3Expr(c_op, lhs <= rhs);
       break;
 
     case clang::BO_EQ:
@@ -546,6 +554,14 @@ bool Z3ConvVisitor::VisitBinaryOperator(clang::BinaryOperator *c_op) {
 
     case clang::BO_NE:
       InsertZ3Expr(c_op, lhs != rhs);
+      break;
+
+    case clang::BO_LAnd:
+      InsertZ3Expr(c_op, Z3BoolCast(lhs) && Z3BoolCast(rhs));
+      break;
+
+    case clang::BO_LOr:
+      InsertZ3Expr(c_op, Z3BoolCast(lhs) || Z3BoolCast(rhs));
       break;
 
     case clang::BO_Rem:
@@ -564,8 +580,20 @@ bool Z3ConvVisitor::VisitBinaryOperator(clang::BinaryOperator *c_op) {
       InsertZ3Expr(c_op, lhs & rhs);
       break;
 
+    case clang::BO_Or:
+      InsertZ3Expr(c_op, lhs | rhs);
+      break;
+
     case clang::BO_Xor:
       InsertZ3Expr(c_op, lhs ^ rhs);
+      break;
+
+    case clang::BO_Mul:
+      InsertZ3Expr(c_op, lhs * rhs);
+      break;
+
+    case clang::BO_Div:
+      InsertZ3Expr(c_op, lhs / rhs);
       break;
 
     case clang::BO_Shr:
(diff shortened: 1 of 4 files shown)


rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -564,18 +580,34 @@ bool Z3ConvVisitor::VisitBinaryOperator(clang::BinaryOperator *c_op) {
...
+      break;
+
     case clang::BO_Shr:
       InsertZ3Expr(c_op, c_op->getLHS()->getType()->isSignedIntegerType()
                              ? z3::ashr(lhs, rhs)
                              : z3::lshr(lhs, rhs));
       break;
 
+    case clang::BO_Shl:
+      InsertZ3Expr(c_op, z3::shl(lhs, rhs));
```
meme: Does this make sense?

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -790,13 +847,36 @@ void Z3ConvVisitor::VisitBinaryApp(z3::expr z_op) {
       }
     } break;
 
+    case Z3_OP_XOR:
```
meme: There is also BXOR, I am not too entirely familiar with Z3, so I assume I want BXOR, which is what https://github.com/lifting-bits/rellic/pull/52/files#diff-0ad1217b54cf74a0a8b80e2ac5ac98b6R588 emits?

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -790,13 +847,36 @@ void Z3ConvVisitor::VisitBinaryApp(z3::expr z_op) {
...
+      break;
+
+    case Z3_OP_BMUL:
+      c_op = CreateBinaryOperator(*ast_ctx, clang::BO_Mul, lhs, rhs,
+                                  GetIntResultType());
+      break;
+
+    case Z3_OP_BSDIV_I:
+    case Z3_OP_BUDIV_I:
+      c_op = CreateBinaryOperator(*ast_ctx, clang::BO_Div, lhs, rhs,
```
meme: Do I want non-`_I` variants here, too? I've been avoiding floats in all my PRs

surovic:
Will keep an eye on this one. I'll get to the reviews by the end of next week.

rellic/AST/Z3ConvVisitor.cpp
surovic: Yeah, I think we can omit floats until someone explicitly needs or wants them in.

rellic/AST/Z3ConvVisitor.cpp
surovic: Yup, `BXOR` is what you want here. The `B` prefix denotes Z3 operations over bitvectors, which you can imagine as C integers of certain width. For example, `uint32_t` would be a `(_ BitVec 32)` in Z3. If you're interested in learning a bit more, you can check out the Z3 [tutorial](https://rise4fun.com/z3/tutorialcontent/guide) and [C++ API](https://z3prover.github.io/api/html/group__cppapi.html).

rellic/AST/Z3ConvVisitor.cpp
surovic: Yeah, I think I would've written the same ;)

meme:
Thanks for the review, I'll get in the BXOR change and make sure I have some better test coverage for these changes

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -548,6 +556,14 @@ bool Z3ConvVisitor::VisitBinaryOperator(clang::BinaryOperator *c_op) {
...
+    case clang::BO_LAnd:
+      InsertZ3Expr(c_op, Z3BoolCast(lhs) && Z3BoolCast(rhs));
+      break;
+
+    case clang::BO_LOr:
+      InsertZ3Expr(c_op, Z3BoolCast(lhs) || Z3BoolCast(rhs));
+      break;
+
     case clang::BO_Rem:
       InsertZ3Expr(c_op, z3::srem(lhs, rhs));
```
surovic: We should differentiate between signed and unsigned remainders here. If the `clang::BO_Rem` operation yields a signed result, we should use `z3::srem()`, `z3::urem()` otherwise.

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -564,18 +580,34 @@ bool Z3ConvVisitor::VisitBinaryOperator(clang::BinaryOperator *c_op) {
...
+
     case clang::BO_Xor:
       InsertZ3Expr(c_op, lhs ^ rhs);
       break;
 
+    case clang::BO_Mul:
+      InsertZ3Expr(c_op, lhs * rhs);
+      break;
+
+    case clang::BO_Div:
```
surovic: See above comment on remainders. `z3::udiv()` and `z3::sdiv()` should be used for signed and unsigned operations, respectively.

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -532,12 +532,20 @@ bool Z3ConvVisitor::VisitBinaryOperator(clang::BinaryOperator *c_op) {
   auto rhs = GetOrCreateZ3Expr(c_op->getRHS());
   // Create z3 binary op
   switch (c_op->getOpcode()) {
-    case clang::BO_LAnd:
-      InsertZ3Expr(c_op, Z3BoolCast(lhs) && Z3BoolCast(rhs));
+    case clang::BO_GT:
```
surovic: We should differentiate between signed and unsigned comparisons here. If `lhs` of the comparison is signed, we should use, for example, `z3::sgt()`, otherwise `z3::ugt()`.

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -772,6 +804,31 @@ void Z3ConvVisitor::VisitBinaryApp(z3::expr z_op) {
                                   ast_ctx->BoolTy);
       break;
 
+    case Z3_OP_SLEQ:
```
surovic: See above comments on signedness. Correct behavior should probably be enforced via `clang::CStyleCastExpr` expressions. This goes for quotient (division), remainder and comparison operations alike.

surovic:
As you'll probably notice, there's a ton of stuff related to signedness in the comments. I've omitted these semantics in the past which is why they're not in the code currently. But I guess this PR is as good as any time to add it. Take your time, no rush.

meme:
Don't have time to work on this unfortunately, hopefully someone (or I) gets the chance to revive it.

PR #63, opened 2020-04-22
Minimize Docker image

ekilmer:
The docker image was previously around 2GB, which is way too big. This PR aims to minimize the Docker image to something a bit more reasonable: 313MB as listed by `docker images` (the compressed image that people would `docker pull` should be a little smaller)

Since I base the final image on `ubuntu:18.04`, we can calculate the size of rellic installation by subtracting the base image size `64.2MB` by the total size of our image for `248.8MB`

Here is a listing of file sizes in the install directory:
```
$ docker build -t rellic .
$ docker run --rm -ti --entrypoint=/bin/bash rellic

root@c8585327edf5:/# du -ha /opt/rellic/
113M    rellic/lib/librellic.a
113M    rellic/lib
36M     rellic/bin/rellic-headergen-8.0
89M     rellic/bin/rellic-decomp-8.0
124M    rellic/bin
4.0K    rellic/docker-decomp-entrypoint.sh
236M    rellic/
```

`236M` doesn't equal `248.8MB`, as I listed above, and I think this is due to the dependency on `libomp5` in the `base` image.

Another thing is I don't think we need to include `librellic.a` in the docker image, which should reduce the size by almost half. @surovic What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,15 @@ jobs:
     - services:
         - docker
       name: 'Docker build'
-      packages:
-        - clang-8
+      addons:
+        apt:
+          sources:
+            - llvm-toolchain-xenial-8
+          packages:
+            - clang-8
       script:
-        - docker build -t rellic-decomp --build-arg LLVM_VERSION=8.0 .
         - clang-8 -emit-llvm -c ./tests/tools/decomp/issue_4.c -o ./tests/tools/decomp/issue_4.bc
+        - docker build -t rellic-decomp --build-arg LLVM_VERSION=8.0 .
         - |
           docker run --rm -t \
             -v "$(pwd)":/test -w /test \
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,11 @@ jobs:
     - services:
         - docker
       name: 'Docker build'
+      packages:
+        - clang-8
       script:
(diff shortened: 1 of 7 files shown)


ekilmer:
For some reason, I'm getting a segfault when running tests for LLVM 7 version of rellic in Docker. Will investigate later.

ekilmer:
Fixed the LLVM 7 error by using the host compiler instead of the downloaded LLVM compiler from `cxxcommon` 🤷 

Also, the total image size is now down to `166MB` with the rellic directory being now `90MB`:
```
root@867b61dde116:/# du -ha /opt/rellic/
30M     /opt/rellic/bin/rellic-headergen-8.0
60M     /opt/rellic/bin/rellic-decomp-8.0
90M     /opt/rellic/bin
4.0K    /opt/rellic/docker-decomp-entrypoint.sh
90M     /opt/rellic/
```

surovic:
>Another thing is I don't think we need to include librellic.a in the docker image, which should reduce the size by almost half. @surovic What do you think?

Agreed.

surovic:
Looks good to me. @artemdinaburg unless you see anything that needs attention, I'm gonna merge o/

artemdinaburg:
Merge it

PR #64, opened 2020-05-14
Use cxx-common docker base

artemdinaburg:
* Base image is cxx-common
* Build Z3 in Dockerfile
* Create a build-only option to avoid downloading cxx-common
* Migrate to cmake -G Ninja versus makefiles

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ jobs:
             - clang-8
       script:
         - clang-8 -emit-llvm -c ./tests/tools/decomp/issue_4.c -o ./tests/tools/decomp/issue_4.bc
-        - docker build . -t rellic:llvm800-ubuntu18.04-amd64 -f Dockerfile --build-arg DISTRO_BASE=ubuntu18.04 --build-arg ARCH=amd64 --build-arg LLVM_VERSION=800
+        - docker build . -t rellic:llvm800-ubuntu18.04-amd64 -f Dockerfile --build-arg UBUNTU_VERSION=18.04 --build-arg ARCH=amd64 --build-arg LLVM_VERSION=800
         - |
           docker run --rm -t \
             -v "$(pwd)":/test -w /test \
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,4 +33,4 @@ jobs:
           docker run --rm -t \
             -v "$(pwd)":/test -w /test \
             -u "$(id -u)":"$(id -g)" \
-            rellic-decomp:latest --input ./tests/tools/decomp/issue_4.bc --output /dev/stdout
+            rellic:llvm800-ubuntu18.04-amd64 --input ./tests/tools/decomp/issue_4.bc --output /dev/stdout
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,31 @@
 # Choose your LLVM version
-ARG LLVM_VERSION=8.0
+ARG LLVM_VERSION=800
+ARG ARCH=amd64
+ARG DISTRO_BASE=ubuntu18.04
+ARG LIBRARIES=/opt/trailofbits/libraries
+ARG Z3_VERSION=4.7.1
+#this seems to be static for most Z3 Linux builds
+ARG Z3_OS_VERSION=ubuntu-16.04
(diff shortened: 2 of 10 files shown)


Dockerfile
```diff
@@ -1,15 +1,32 @@
...
+ARG DISTRO_BASE=ubuntu18.04
+ARG LIBRARIES=/opt/trailofbits/libraries
+ARG Z3_VERSION=4.7.1
+#this seems to be static for most Z3 Linux builds
+ARG Z3_OS_VERSION=ubuntu-16.04
+ARG Z3_ARCHIVE=z3-${Z3_VERSION}-x64-${Z3_OS_VERSION}
 
 # Run-time dependencies go here
-FROM ubuntu:18.04 as base
+FROM trailofbits/cxx-common:llvm${LLVM_VERSION}-${DISTRO_BASE}-${ARCH} as base
```
ekilmer, line 12: Using `cxx-common` as a base here will result in a very large final image, since `cxx-common` is huge.

What is the end-goal of this Docker image? I thought we wanted to just package `rellic` by itself in the resulting Docker image

Dockerfile
```diff
@@ -1,15 +1,32 @@
...
     rm -rf /var/lib/apt/lists/*
 
+#TODO: replace with git clone & build of the version to work on non-amd64 arch
+RUN curl -OL "https://github.com/Z3Prover/z3/releases/download/z3-${Z3_VERSION}/${Z3_ARCHIVE}.zip" && \
+    unzip -qq "${Z3_ARCHIVE}.zip" && \
+    mv "${Z3_ARCHIVE}" "${LIBRARIES}/z3" && \
+    rm "${Z3_ARCHIVE}.zip"
+
 # Build-time dependencies go here
 FROM base as deps
```
ekilmer, line 31: We could think about putting `cxx-common` here, as part of the build dependencies, which won't be included in the final image.

Dockerfile
artemdinaburg: Wouldn't we need cxx-common to run it?

artemdinaburg:
Build example:
```docker build . -t rellic:llvm800-ubuntu18.04-amd64      -f Dockerfile      --build-arg DISTRO_BASE=ubuntu18.04      --build-arg ARCH=amd64      --build-arg LLVM_VERSION=800```

**Comment by ekilmer** [comment]
@artemdinaburg I reworked a few more lines to make sure it builds on ubuntu 16.04, 18.04, 19.10, and 20.04.

We might want to add all the Ubuntu versions that test to the CI, but let's wait until we transition to Github Actions to do that, I think.

File: .travis.yml [hunk 3]
```diff
@@ -28,7 +28,7 @@ jobs:
             - clang-8
       script:
         - clang-8 -emit-llvm -c ./tests/tools/decomp/issue_4.c -o ./tests/tools/decomp/issue_4.bc
-        - docker build . -t rellic:llvm800-ubuntu18.04-amd64 -f Dockerfile --build-arg DISTRO_BASE=ubuntu18.04 --build-arg ARCH=amd64 --build-arg LLVM_VERSION=800
+        - docker build . -t rellic:llvm800-ubuntu18.04-amd64 -f Dockerfile --build-arg UBUNTU_VERSION=18.04 --build-arg ARCH=amd64 --build-arg LLVM_VERSION=800
```
ekilmer, line 31: Using the build-arg `UBUNTU_VERSION` should be the preferred way now, since we use that to test how we install Python 3.7 (or later)

scripts/roundtrip.py
```diff
@@ -1,4 +1,24 @@
...
+
+echo "Python not found, exiting!" >2
+
+exit 2
+
+":"""
+# Previous line is bilingual: it ends a comment in Python, and is a no-op in shell
+# Shell commands end here
+# Python script follows
+# Credit for above: https://stackoverflow.com/a/47886254
```
ekilmer, line 21: This is really painful, and I can remove/change it to something else if there's a better option.

The bad thing about this way is that if python3.9 comes out, it will have to be updated, but then I hope it will be caught by the `python3` option

.travis.yml
ekilmer: No longer relevant to Python.

scripts/roundtrip.py
ekilmer: Removed.

PR #93, opened 2021-02-02
vcpkg: Support Windows and arm64

ekilmer:
Required for https://github.com/trailofbits/cxx-common/pull/677 and https://github.com/trailofbits/cxx-common/pull/679


pgoodman (approved):
Is the z3 complexity here to support non-vcpkg builds? Will this work?

```
find_package(Z3 CONFIG REQUIRED)
add_library(Z3 INTERFACE)
target_link_libraries(Z3 INTERFACE z3::libz3)
```

(This is how, on macOS at least, I got around LLVM needing `-lZ3` but then `z3::libz3` introducing `-llibz3.a`)

ekilmer:
Good question. The Z3 complexity is necessary because I forgot to exclude the line in the CMakeLists.txt for vcpkg builds here https://github.com/lifting-bits/rellic/blob/bc53135a4c4697c600fd27f2013ef1d66d15ad24/CMakeLists.txt#L27

That line tells CMake to first look at our local `FindZ3.cmake`, which is a little bugged for Windows. It's fixed for each LLVM in this PR https://github.com/trailofbits/cxx-common/pull/677/files#diff-d291b108800f56606be4b9351a183391e21f44d6be13e2f1199c467c2451a23b

I'm going to revert the changes to this repo's `FindZ3.cmake` and add a conditional to the line above.

PR #117, opened 2021-03-05
Update tests to add challenge1

artemdinaburg:
* Add challenge-1 as a test
* Update tests to handle things that should translate but not rebuild
* Use --lower_switch and --remove_phi_nodes by default


artemdinaburg:
Closing this since this PR has been merged into #119

PR #119, opened 2021-03-09
Implements a proper z3::extract translation

### LINKED ISSUE #87: Over-eager optimizations
Some of our optimizations are too eager and remove some needed math.

For example, something like:
```
  uint8_t brake_switch = (buf[4] & 0b00001100) >> 2;
...
  if (brake_switch){
...
```

Will decompile to:
```
if (((arg0[4U])) != '\x00') {
```

This is missing come critical shifts and bitops in the conditional.

surovic:
Aside from the subject matter, it moves around some functions and makes a few handlers simpler thanks to it.

Resolves #87 

NOTE: The translation visitor are getting way too bloated and it's starting to impact time it takes to do anything. #6 will have to happen pretty soon.


rellic/AST/Util.cpp
```diff
@@ -68,6 +68,45 @@ bool ReplaceChildren(clang::Stmt *stmt, StmtMap &repl_map) {
...
+  auto src = op->getType();
+  // Helpers
+  auto IsInt = [&ctx](clang::QualType t) { return t->isIntegralType(ctx); };
+  auto IsFloat = [](clang::QualType t) { return t->isFloatingType(); };
+  auto IsSmaller = ctx.getTypeSize(src) < ctx.getTypeSize(dst);
+  auto MakeCast = [&ctx, dst, op](clang::CastKind kind) {
+    return CreateCStyleCastExpr(ctx, dst, kind, op);
+  };
+
+  // CK_FloatingCast
```
pgoodman, line 90: Can you change these comments to describe operation being detected/handled? E.g, float to float and issmaller could be described as "Widen the floating point type, e.g. `fpext` in LLVM bitcode."

What happens if source and dest operand types are different, but the size is the same? e.g. `int64_t` to `double`, or `int32_t` to `float`?

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -202,25 +202,28 @@ clang::Expr *Z3ConvVisitor::CreateLiteralExpr(z3::expr z_expr) {
...
-      auto &ti{ast_ctx->getTargetInfo()};
-      auto t_size{ti.getTypeWidth(ti.getLeastIntTypeByWidth(s_size, 0))};
-      auto t_lit{ast_ctx->getIntTypeForBitwidth(t_size, 0)};
-      llvm::APInt val(t_size, Z3_get_numeral_string(z_expr.ctx(), z_expr), 10);
-      switch (clang::cast<clang::BuiltinType>(t_lit)->getKind()) {
+      auto type{GetLeastIntTypeForBitWidth(*ast_ctx, GetZ3SortSize(sort),
+                                           /*sign=*/0)};
+      auto size{ast_ctx->getTypeSize(type)};
+      llvm::APInt val(size, Z3_get_numeral_string(z_expr.ctx(), z_expr), 10);
+      switch (clang::cast<clang::BuiltinType>(type)->getKind()) {
```
pgoodman, line 209: Can you add a comment here about why `ushort` and `uchar` need special handling, but others do not?

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -794,14 +797,31 @@ void Z3ConvVisitor::VisitUnaryApp(z3::expr z_op) {
...
-      auto &ti{ast_ctx->getTargetInfo()};
-      auto t_size{ti.getTypeWidth(ti.getLeastIntTypeByWidth(s_size, t_sign))};
-      auto t_op{ast_ctx->getIntTypeForBitwidth(t_size, t_sign)};
-      c_op = CreateCStyleCastExpr(*ast_ctx, t_op,
-                                  clang::CastKind::CK_IntegralCast, c_sub);
+      auto t_uint{ast_ctx->UnsignedIntTy};
+      auto t_uint_size{ast_ctx->getTypeSize(t_uint)};
+      auto t_res{ast_ctx->getIntegerTypeOrder(t_sub, t_uint) < 0 ? t_uint
+                                                                 : t_sub};
+      // Shift value
```
pgoodman, line 804: Can you change these comments to describe the higher level transformation you're doing? Somthing like, `Given a `(extract blah A B)` produce an expression of the form `(blah >> foo) & bar`." Right now the commenting here provides no additional value over your variable names.

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -869,38 +896,37 @@ void Z3ConvVisitor::VisitBinaryApp(z3::expr z_op) {
...
+      c_op = BinOpExpr(clang::BO_EQ, ast_ctx->IntTy);
       break;
 
     case Z3_OP_SLEQ:
-      c_op = CreateBinaryOperator(*ast_ctx, clang::BO_LE, lhs, rhs,
-                                  ast_ctx->IntTy);
+      c_op = BinOpExpr(clang::BO_LE, ast_ctx->IntTy);
       break;
 
     case Z3_OP_AND: {
```
pgoodman, line 906: Can you add a comment here that says something like `AND`s in z3 can be variadic, and so you lower / fold them into something like `(A & (B & C))` or `((A & B) & C)`, and make a note of `lhs` and `rhs` being shared byref with `BinOpExpr`.

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -869,38 +896,37 @@ void Z3ConvVisitor::VisitBinaryApp(z3::expr z_op) {
...
-        c_op = CreateBinaryOperator(*ast_ctx, clang::BO_LAnd, c_op, rhs,
-                                    ast_ctx->IntTy);
+        lhs = BinOpExpr(clang::BO_LAnd, ast_ctx->IntTy);
       }
+      c_op = lhs;
     } break;
 
     case Z3_OP_OR: {
-      c_op = lhs;
       for (auto i = 1U; i < z_op.num_args(); ++i) {
```
pgoodman, line 915: Ditto for here as with ANDs.

rellic/AST/Z3ConvVisitor.cpp
```diff
@@ -869,38 +896,37 @@ void Z3ConvVisitor::VisitBinaryApp(z3::expr z_op) {
...
     case Z3_OP_CONCAT: {
-      if (auto c_lit = clang::dyn_cast<clang::IntegerLiteral>(lhs)) {
+      auto c_lit{
+          clang::dyn_cast<clang::IntegerLiteral>(lhs->IgnoreParenCasts())};
+      if (c_lit && (c_lit->getValue().isNullValue() ||
+                    c_lit->getValue().isAllOnesValue())) {
         auto val = c_lit->getValue();
         auto s_size = GetZ3SortSize(z_op.get_sort());
-        auto t_sign = val.isNullValue() ? 0 : 1;
+        auto t_sign = val == 0 ? 0U : 1U;
```
pgoodman, line 929: Can you add a comment here describing how signedness is inferred or communicated here?

pgoodman (changes requested):

artemdinaburg:
Bitcode of failing MacOS test attached
[assert.bc.gz](https://github.com/lifting-bits/rellic/files/6110868/assert.bc.gz)

rellic/AST/Util.cpp
surovic: >What happens if source and dest operand types are different, but the size is the same? e.g. int64_t to double, or int32_t to float?

This should be covered by the latter two cases. Just a simple `IntegralToFloating` cast.

rellic/AST/Z3ConvVisitor.cpp
pgoodman: Are there any possible downstream interactions with the order of the `&&` or `||` folding? I.e. if it's `(A && (B && ...))` vs. `((...) && C)`, will the short-circuiting of `C` interact in any way? I guess it depends on if a result of an effectful result is stored to something like `C`, and making sure it's not inlined in place of `C`.

pgoodman (approved):

artemdinaburg (approved):

PR #120, opened 2021-03-11
Skip annotation intrinsics when translating to C

artemdinaburg:
* Skip annotation intrinsic functions
* Skip global variables that go into metadata like llvm.used

### DIFF
--- a/rellic/AST/IRToASTVisitor.cpp
+++ b/rellic/AST/IRToASTVisitor.cpp
@@ -240,8 +240,10 @@ clang::Expr *IRToASTVisitor::CreateLiteralExpr(llvm::Constant *constant) {
   return result;
 }
 
-#define ASSERT_ON_VALUE_TYPE(x) \
-  if (llvm::isa<x>(val)) { LOG(FATAL) << "Invalid operand [" #x "]"; }
+#define ASSERT_ON_VALUE_TYPE(x)               \
+  if (llvm::isa<x>(val)) {                    \
+    LOG(FATAL) << "Invalid operand [" #x "]"; \
+  }
 
 clang::Expr *IRToASTVisitor::GetOperandExpr(llvm::Value *val) {
   DLOG(INFO) << "Getting Expr for " << LLVMThingToString(val);
@@ -415,7 +417,7 @@ void IRToASTVisitor::VisitFunctionDecl(llvm::Function &func) {
   auto name = func.getName().str();
   DLOG(INFO) << "VisitFunctionDecl: " << name;
 
-  if(IsAnnotationIntrinsic(func.getIntrinsicID())) {
+  if (IsAnnotationIntrinsic(func.getIntrinsicID())) {
     DLOG(INFO) << "Skipping creating declaration for LLVM intrinsic";
     return;
   }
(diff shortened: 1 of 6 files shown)


rellic/AST/IRToASTVisitor.cpp
```diff
@@ -435,6 +448,30 @@ void IRToASTVisitor::VisitFunctionDecl(llvm::Function &func) {
...
+  // NOTE(artem): As of this writing, rellic does not do anything
+  // with debug intrinsics and debug metadata. Processing them to C
+  // is useless (since there is no C equivalent anyway).
+  // All it does it lead to triggering of asserts for things that are
+  // low-priority on the "to fix" list
+
+  if (llvm::isDbgInfoIntrinsic(inst.getIntrinsicID())) {
+    DLOG(INFO) << "Skipping debug data intrinsic";
+    return;
+  } else if (IsAnnotationIntrinsic(inst.getIntrinsicID())) {
```
surovic, line 463: We can skip the `else` here, since the `if` block returns.

rellic/AST/IRToASTVisitor.cpp
```diff
@@ -899,8 +936,14 @@ void IRToASTVisitor::visitCastInst(llvm::CastInst &inst) {
   if (cast) {
     return;
   }
+
+  auto oper = inst.getOperand(0);
```
surovic, line 940: A cast without an operand? What triggers this?

surovic (changes requested):

rellic/AST/IRToASTVisitor.cpp
artemdinaburg: This was here from debugging some issues

rellic/AST/IRToASTVisitor.cpp
artemdinaburg: Fixed

rellic/AST/IRToASTVisitor.cpp
artemdinaburg: Fixed

rellic/BC/Util.h
```diff
@@ -37,4 +37,11 @@ bool VerifyModule(llvm::Module *module);
 llvm::Module *LoadModuleFromFile(llvm::LLVMContext *context,
                                  std::string file_name,
                                  bool allow_failure = false);
+
+// Check if an intrinsic ID is an annotation
+bool IsAnnotationIntrinsic(llvm::Intrinsic::ID id);
```
pgoodman, line 42: `llvm::Intrinsic::ID` is not declared in this header. It is a `typedef unsigned ID;` in the `llvm::Intrinsic` namespace. Forward declare with:

```c++
namespace Intrinsic {
typedef unsigned ID;
}  // namespace Instrinsic
```
up above in the LLVM namespace.

rellic/BC/Util.h
```diff
@@ -37,4 +37,11 @@ bool VerifyModule(llvm::Module *module);
...
+
+// check if a global object is llvm metadata
+bool IsGlobalMetadata(const llvm::GlobalObject &go);
```
pgoodman, line 45: Forward declare the `GlobalObject` class in the LLVM namespace above.

rellic/BC/Compat/IntrinsicInst.h
```diff
@@ -0,0 +1,17 @@
+#include "rellic/BC/Version.h"
+
```
pgoodman, line 2: Include the LLVM `IntrinsicInst.h` header here, so that in place of including LLVM's intrinsic inst header, this header always gets included.

pgoodman (changes requested):

pgoodman (approved):

PR #135, opened 2021-05-07
Run AnghaBench small runs in CI

artemdinaburg:
* Do small anghabench runs in CI via cron and on PRs


surovic (changes requested):
Can we move `run-on-anghabench.sh` to `script` in the root of the repo? Also move the CI submodule to a `libraries` directory under the root. I have `doctest` there in the `issue_6_ast_builder` branch as a submodule too. Aside from that, LGTM.

The tests directory should really mimic the directory structure of `tools` and `lib` and should contain integration/regression test cases.

surovic (approved):

PR #151, opened 2021-07-01
Fixpoint expression folding and fixing expression precedence 

surovic:
Taking suggestions on the janky "precedence table" thing.


rellic/AST/ASTBuilder.cpp
```diff
@@ -18,6 +18,74 @@
 
 namespace rellic {
 
+namespace {
```
pgoodman, line 21: Is it possible for this code to produce an operator precedence that differs from what is already in the AST? What do the numbers `0u`, `1u`, and `2u` mean? Can you maybe make some constexprs or describe in a comment here the rough idea, e.g. 0 is lower precedence, or 0 is higher precedence, etc.

pgoodman: Also, how is associativity managed?

tools/decomp/Decomp.cpp
```diff
@@ -126,22 +126,27 @@ static bool GeneratePseudocode(llvm::Module& module,
...
-      z3::tactic(fin_simplifier->GetZ3Context(), "solve-eqs") &
+      // Cheap simplification
+      z3::tactic(fin_simplifier->GetZ3Context(), "simplify") &
       // Propagate bounds over bit-vectors
       z3::tactic(fin_simplifier->GetZ3Context(), "propagate-bv-bounds") &
       // Contextual simplification
       z3::tactic(fin_simplifier->GetZ3Context(), "ctx-simplify"));
 
-  llvm::legacy::PassManager fin;
+  llvm::legacy::PassManager scope_folder;
```
pgoodman, line 136: Maybe move this declaration up above the `auto fin_simplifier{...`, and then make the creation of `fin_simplifier` conditional on `FLAGS_disable_z3`, so that we don't leak the `Z3CondSimplify` when Z3 isn't used.

pgoodman (changes requested):

rellic/AST/ASTBuilder.cpp
surovic: > Is it possible for this code to produce an operator precedence that differs from what is already in the AST?
Yes. A common problem I had were ASTs like `(subscript (+ ptr offset) index)`, which would be printed as `ptr + offset[index]`. This is wrong of course and should be `(ptr + offset)[index]`.

> What do the numbers 0u, 1u, and 2u mean?
These are a technicality that I need to properly create an "ordering" of the C operators by precedence. There's `enum clang::UnaryOperatorKind` which _I think_ goes from 0. Same for `enum clang::BinaryOperatorKind`. These both contain the majority of unary and binary operators of C and are internally ordered by precedence starting from highest priority to lowest.

Then there's operators that are their own class, like `clang::MemberExpr` which corresponds to `.` and `->` operators and `clang::ArraySubscriptExpr` which corresponds to `[]`. Finally there's declaration references and literals, which have the highest "precedence".

So finally, to get an ordering that would correspond to the correct precedence from highest to lowest I give declaration references and literals precedence `0U` (highest). After that, `1U` are parenthesized expressions, member, array subscript, call and type casting expressions. And finally `2U` and beyond are members of `clang::UnaryOperatorKind`, `2U + idx-of-lowest-unary-operator` and beyond are members of `clang::BinaryOperatorKind` and beyond that is `clang::ConditionalOperator`.

Admittedly it's not the most elegant solution, but it seems to work. I'm _very_ open to better solutions.

> Also, how is associativity managed?
My hunch is that associativity is taken care of implicitly by how the AST is constructed and printed.

pgoodman (approved):

rellic/AST/ASTBuilder.cpp
pgoodman: Thanks for explaining!

PR #158, opened 2021-07-14
Fix handling of undef integer values.

artemdinaburg:
Fix handling of `undef` integer values
This was the source of a segfault due to a failed cast.

### DIFF
--- a/rellic/AST/IRToASTVisitor.cpp
+++ b/rellic/AST/IRToASTVisitor.cpp
@@ -159,12 +159,7 @@ clang::Expr *IRToASTVisitor::CreateLiteralExpr(llvm::Constant *constant) {
           result = ast.CreateAdjustedIntLit(val);
         }
       } else if (llvm::isa<llvm::UndefValue>(constant)) {
-        //NOTE(artem):
-        // This gives the following result:
-        // warning: indirection of non-volatile null pointer will be deleted, not trap [-Wnull-dereference]
-        // note: consider using __builtin_trap() or qualifying pointer with 'volatile'
-        // We may want to consider making a CreteTrap() instead?
-        result = ast.CreateUndef(c_type);
+        result = ast.CreateUndefInteger(c_type);
       } else {
         LOG(FATAL) << "Unsupported integer constant";
       }
@@ -174,7 +169,7 @@ clang::Expr *IRToASTVisitor::CreateLiteralExpr(llvm::Constant *constant) {
       if (llvm::isa<llvm::ConstantPointerNull>(constant)) {
         result = ast.CreateNull();
       } else if (llvm::isa<llvm::UndefValue>(constant)) {
-        result = ast.CreateUndef(c_type);
+        result = ast.CreateUndefPointer(c_type);
       } else {
         LOG(FATAL) << "Unsupported pointer constant";
       }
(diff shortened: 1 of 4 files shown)


rellic/AST/IRToASTVisitor.cpp
```diff
@@ -150,11 +150,23 @@ clang::Expr *IRToASTVisitor::CreateLiteralExpr(llvm::Constant *constant) {
...
+      if (llvm::isa<llvm::ConstantInt>(constant)) {
+        auto ci = llvm::cast<llvm::ConstantInt>(constant);
+        auto val{ci->getValue()};
+        if (val.getBitWidth() == 1U) {
+          result = ast.CreateIntLit(val);
+        } else {
+          result = ast.CreateAdjustedIntLit(val);
+        }
+      } else if (llvm::isa<llvm::UndefValue>(constant)) {
+        //NOTE(artem):
```
pgoodman, line 162: This is probably going to be a more general issue, of having to deal with `undef` vals. I'm going to wait for @surovic to implement a fix in a way that he sees fit, as he might want to factor out handling of `undef`s.

surovic: I don't really have a plan for `undef` values yet. So for now, I guess this is a good enough fix. Added as issue #159 

@artemdinaburg do you have an input that I could put to the issue?

artemdinaburg: So I had to read the [undefined value description](https://llvm.org/docs/LangRef.html#undefvalues) a bit more. 

An `undef` (if not frozen, which we have not run into yet) can take *any* value. This is not undefined behavior, but an undefined value.

My opinion is that:

1) This is probably due to some anvill issue, and anvill should definitely have a pass to see if `undef` leaks out to final bitcode.
2) We may have `undef` as a legitimate source into rellic, if, say, the ISA specs for some instruction specify a register is in an undefined state, and then the register is used.
3) The current behavior (assume that an `undef` pointer is NULL) is perfectly valid, since it can indeed be NULL. 
4) For this specific PR, we should return a constant integer of 0 of the right bit width (not cast to a pointer, as this is an int value). I will see if I can make the change.
5) From a correctness and "what does the user expect?" standpoint, I think we should pick a value (say, zero) and emit function call like:

```
int32_t rellic_undefined_value_int32() {
#pragma error "The input bitcode uses an undefined value. This may not be an error, but it probably is. Uncomment this pragma to achieve recompilable code.
return 0; // any value works here
}

...
int var0 = rellic_undefined_value_int32();
...

```
The user of the decompiled code can then figure out what they want to do about the `undef`, and static analysis tools should be trivially able to see that `rellic_undefined_value_*()` returns 0 (or 42 or whatever we pick).

unittests/AST/ASTBuilder.cpp
```diff
@@ -267,7 +267,25 @@ TEST_SUITE("ASTBuilder::CreateNull") {
   }
 }
 
-TEST_SUITE("ASTBuilder::CreateUndef") {
+TEST_SUITE("ASTBuilder::CreateUndefInteger") {
```
surovic, line 270: Can you merge `TEST_SUITE("ASTBuilder::CreateUndefInteger")` and `TEST_SUITE("ASTBuilder::CreateUndefPointer")` into `TEST_SUITE("ASTBuilder::CreateUndef*")` and have them share some of the common code?

surovic (changes requested):

pgoodman (approved):

surovic (approved):

PR #172, opened 2021-08-25
Implements a custom AST printer

surovic:
The printer largely uses the default `clang::Stmt::printPretty`, `clang::Decl::print()` and `clang::QualType::print` methods but with changes to how `clang::FunctionDecl` is printed.


pgoodman:
Can we merge this?

PR #174, opened 2021-09-02
Fix missing dependencies in Linux build example

greenbagels:
From the commit message:

> The `doctest-dev` package available in Debian and Ubuntu repositories
was not recently listed in the README.md build instructions for Linux,
but is required. I've added it to the example, as well as switched the
`apt-get` invocations to the canonically-preferred `apt` tool.

I'm a first-time contributor to this project, so I don't know if anyone has
strong opinions on the `apt-get` vs `apt` thing; since the example seems
more oriented towards interactive usage rather than being in a script, I
figured the change might be an apt choice :)

If people *do* think it's unnecessary, I'm more than happy to undo the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,10 @@ Pre-built Docker images are available on [Docker Hub](https://hub.docker.com/rep
 First, update aptitude and get install the baseline dependencies.
 
 ```shell
-sudo apt-get update
-sudo apt-get upgrade
+sudo apt update
+sudo apt upgrade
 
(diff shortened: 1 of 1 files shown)


greenbagels:
Also a minor typo that I'll squash into the commit in a bit: "resuling"

https://github.com/lifting-bits/rellic/blob/d451245e3043981c65ea3d09019c3eef7d312a5b/README.md#L134

pgoodman (approved):

PR #176, opened 2021-09-10
Misc fixes

surovic:
* Changes handling of `NullToPointer` casts to be the same as `IntegralToPointer` in C -> Z3 codegen
* Adds C codegen for `freeze` LLVM IR instruction
* Adds C codegen for `i128` LLVM IR constants

### DIFF
--- a/lib/AST/IRToASTVisitor.cpp
+++ b/lib/AST/IRToASTVisitor.cpp
@@ -166,7 +166,9 @@ clang::Expr *IRToASTVisitor::CreateLiteralExpr(llvm::Constant *constant) {
         if (val_bitwidth == 1U) {
           // Booleans
           result = ast.CreateIntLit(val);
-        } else if (val_bitwidth > ull_bitwidth) {
+        } else if (val.getActiveBits() <= ull_bitwidth) {
+          result = ast.CreateAdjustedIntLit(val);
+        } else {
           // Values wider than `long long` will be represented as:
           // (uint128_t)hi_64 << 64U | lo_64
           auto lo{ast.CreateIntLit(val.extractBits(64U, 0U))};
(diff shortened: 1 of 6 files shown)


lib/AST/IRToASTVisitor.cpp
```diff
@@ -162,8 +162,20 @@ clang::Expr *IRToASTVisitor::CreateLiteralExpr(llvm::Constant *constant) {
       if (llvm::isa<llvm::ConstantInt>(constant)) {
         auto ci{llvm::cast<llvm::ConstantInt>(constant)};
         auto val{ci->getValue()};
-        if (val.getBitWidth() == 1U) {
+        auto val_bitwidth{val.getBitWidth()};
+        auto ull_bitwidth{ast_ctx.getIntWidth(ast_ctx.LongLongTy)};
+        if (val_bitwidth == 1U) {
+          // Booleans
           result = ast.CreateIntLit(val);
+        } else if (val_bitwidth > ull_bitwidth) {
```
pgoodman, line 170: Are you able to check if the actual effective width is <= 64 (there is an `APInt` API for figuring out how many bits are needed to represent the value) and if so, turn it into `(uint128_t) constant`? Possible you can try to also detect if it is a sign-extended 64-bit value, or a `-1` or a `~0`, and then finally, if all else fails, I think the equation you want is:

```
(((uint128_t) high_64) << 64u) | low_64
```

pgoodman (changes requested):

lib/AST/IRToASTVisitor.cpp
surovic: > Are you able to check if the actual effective width is <= 64 (there is an APInt API for figuring out how many bits are needed to represent the value) and if so, turn it into (uint128_t) constant?

I think we can do this. I actually thought we already do.

> Possible you can try to also detect if it is a sign-extended 64-bit value, or a -1 or a ~0

I could try to find bit patterns, but that would mean I'd be chasing special cases. I'd rather have a general solution here and maybe do a rewriting pass later.

> and then finally, if all else fails, I think the equation you want is ...

Yeah, on second thought this is going to be better.  I thought the `udiv` and `urem` would be more universal, but it only really scales towards `i128` anyway.

pgoodman (approved):

PR #179, opened 2021-10-05
Implements a custom AST tokenizer and printer

surovic:


lib/AST/ASTPrinter.cpp
```diff
@@ -0,0 +1,659 @@
...
+    out.push_back(Token::CreateMisc(scs_str));
+    out.push_back(Token::CreateSpace());
+  }
+
+  switch (vdecl->getTSCSpec()) {
+    case clang::TSCS_unspecified:
+      break;
+
+    case clang::TSCS___thread:
+      out.push_back(Token::CreateMisc("__thread"));
```
pgoodman, line 93: Can you attach the `vdecl` here and in the misc stuff below? I'd go as far as even including it on the spaces if that can be done.

lib/AST/ASTPrinter.cpp
```diff
@@ -0,0 +1,659 @@
...
+  if (auto init = vdecl->getInit()) {
+    if (vdecl->getInitStyle() == clang::VarDecl::CInit) {
+      out.push_back(Token::CreateSpace());
+      out.push_back(Token::CreateMisc("="));
+      out.push_back(Token::CreateSpace());
+    } else {
+      LOG(FATAL) << "Invalid initialization style.";
+    }
+
+    StmtTokenizer(out, unit).Visit(init);
```
pgoodman, line 129: Is it possible to associate the `vdecl` with all tokens introduced by this `stmt` (assuming those tokens aren't already tied to another decl)?  In that case, the nodes are associated with a decl and a stmt.

lib/AST/ASTPrinter.cpp
```diff
@@ -0,0 +1,659 @@
...
+  llvm::raw_string_ostream ss(buf);
+  vtype.print(ss, unit.getASTContext().getPrintingPolicy(), vdecl->getName(),
+              /*Indentation=*/0U);
+
+  out.push_back(Token::CreateDecl(vdecl, ss.str()));
+
+  if (auto init = vdecl->getInit()) {
+    if (vdecl->getInitStyle() == clang::VarDecl::CInit) {
+      out.push_back(Token::CreateSpace());
+      out.push_back(Token::CreateMisc("="));
```
pgoodman, line 123: Similar level of decl-assignment here, so that we can really associate all of the tokens to the nodes which generate them.

pgoodman (changes requested):
The tl;dr of my requested changes is to be extremely aggressive and granular about assigning the decl/stmt metadata to tokens. E.g. punctuation and whitespace should have the info associated.

lib/AST/ASTPrinter.cpp
pgoodman: > No, this is not directly possible. 

Can you store the current index into the token output list, then iterate over the tokens added into the list by the call to `Visit`, then assign `vdecl`? This type of thing could probably wrapped in some kind of context manager. If you did the 1:N mapping, then you could have the associated modelled as a vector (a stack really) and then that would naturally coincide with the recursive calls to `.Visit(...)`.

Perhaps even doing everything via context management may prove the easiest solution, because you wouldn't need to manually assign decls/stmts/types to tokens!!

PR #180, opened 2021-10-06
Use debug metadata in LLVM bitcode to infer names

frabert:
I'm creating this draft to make it easier to review code gradually


pgoodman (changes requested):

frabert:
I think I've addressed most of the points with the latest commit

Re: minitool: @surovic Have you already checked if clang-tidy has rules for what you have in mind? It might be especially useful for superfluous includes

Issue #190, opened 2021-10-21
Recover type information


frabert:
Currently, some type information (e.g. `const`, `volatile`) regarding variables is lost.
For example, this source code

```c
struct foo {
  int x;
};

typedef struct foo foo_t;

int main(void) {
  const volatile foo_t **a = 0;
  void *b = a;
}
```

roundtrips to:

```c
unsigned int main();
struct struct_foo {
    unsigned int x;
};
unsigned int main() {
    struct struct_foo **a; // lost const volatile
    unsigned char *b; // was void* before
    a = (void *)0U;
    b = (unsigned char *)a;
    return 0U;
}
```

It would be nice to recover the original type information when possible.

pgoodman:
Do you see `"omnipotent char"` show up anywhere in debug info? We might be able to see that and use `char`. There is [`CharIsSigned`](https://code.woboq.org/llvm/clang/include/clang/Basic/LangOptions.def.html#clang::LangOptionsBase::CharIsSigned), and [here](https://code.woboq.org/llvm/lldb/source/Plugins/TypeSystem/Clang/TypeSystemClang.cpp.html#530) is how LLDB infers this.

frabert:
Yes, `"omnipotent char"` does show up as part of TBAA attributes on `load` and `store` instructions.
I've tried searching for something but couldn't really grasp what that means.

pgoodman:
Perhaps it's actually a sort of joke type for pointers. Perhaps if we know if it's meant to be a `char`, but want to hedge against signed/unsigned, we could have a typedef like `typedef signed char char_t` or `typedef unsigned char char_t`, or have a `char_t` and a `uchar_t`.

PR #191, opened 2021-10-22
Refine types based on debug metadata

frabert:
Solves #190

### DIFF
--- a/lib/AST/IRToASTVisitor.cpp
+++ b/lib/AST/IRToASTVisitor.cpp
@@ -870,8 +870,8 @@ void IRToASTVisitor::visitBinaryOperator(llvm::BinaryOperator &inst) {
   auto rhs{GetOperandExpr(inst.getOperand(1))};
   // Sign-cast int operand
   auto IntSignCast{[this](clang::Expr *operand, bool sign) {
-    auto type{ast_ctx.getIntTypeForBitwidth(
-        ast_ctx.getTypeSize(operand->getType()), sign)};
+    auto type{ast.GetIntTypeForBitWidth(ast_ctx.getTypeSize(operand->getType()),
+                                        sign)};
     return ast.CreateCStyleCast(type, operand);
   }};
   // Where the magic happens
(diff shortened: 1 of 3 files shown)


lib/AST/StructFieldRenamer.cpp
```diff
@@ -60,6 +60,9 @@ bool StructFieldRenamer::VisitRecordDecl(clang::RecordDecl *decl) {
     // FIXME(frabert): Is a clash between field names actually possible?
     // Can this mechanism actually be left out?
     auto name{di_field->getName().str()};
+    if (di_field->getTag() == llvm::dwarf::DW_TAG_inheritance) {
```
pgoodman, line 63: Does this make an explicit field out of the base type in the case of inheritance? Can you add a comment here that shows what a simple c++ code would look like, and what we would generate as a result?

lib/AST/StructFieldRenamer.cpp
```diff
@@ -60,6 +60,9 @@ bool StructFieldRenamer::VisitRecordDecl(clang::RecordDecl *decl) {
...
+      name = di_field->getBaseType()->getName().str() + "_base";
+    }
     if (seen_names.find(name) == seen_names.end()) {
```
pgoodman, line 66: What if you have `seen_names` be a map of `seen_names -> unsinged`, then you could have:

```c++
auto &name_count = seen_names[name];
if (name_count) {
  name = name + "_" + std::to_string(name_count);
}
++name_count;
```

pgoodman (changes requested):

lib/AST/StructFieldRenamer.cpp
pgoodman: Can you add some examples that use multiple inheritance and virtual inheritance?

```c++
struct Base1 {
  int foo;
};
struct Base2 {
  float bar;
}
struct Derived : Base1 , Base2 {
};
```

```c++
struct Base1 {
  int foo;
};
struct Base2 : Base1 {
  float bar;
};
struct Base3 : Base1 {
  float bar;
};
struct Derived : virtual Base2 , virtual Base3 {
};
```

Also, here is a particularly thorny example which shows when this method of embedding the base within the structure of the parent is going to break down:
C++: https://godbolt.org/z/bM4vrq6fW
C: https://godbolt.org/z/fYarYo5he

See this SO post for more detail: https://stackoverflow.com/questions/52818411/will-the-padding-of-base-class-be-copied-into-the-derived-class

lib/AST/IRToASTVisitor.cpp
```diff
@@ -107,11 +114,21 @@ clang::QualType IRToASTVisitor::GetQualType(llvm::Type *type,
 
     case llvm::Type::FunctionTyID: {
       auto func{llvm::cast<llvm::FunctionType>(type)};
-      auto ret{GetQualType(func->getReturnType(),
-                           dic.GetIRFuncTypeToDIRetTypeMap()[func])};
+      std::vector<llvm::DIType *> ditype_array{func->getNumParams() + 1};
```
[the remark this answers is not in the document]
pgoodman: We might need to revisit whether or not it's safe to actually use `this` as an explicit parameter name if we emit c++ code.

PR #197, opened 2021-11-03
Fix ByVal args handling

### LINKED ISSUE #195: Incorrect result when decompiling pass-by-value struct arguments via WebAssembly
Compiling the following program to WebAssembly (no codegen, only IR `clang --target=wasm32-unknown-wasi -S -emit-llvm  ../test_struct_split.c`)
```c
extern int printf(const char *fmt, ...);
extern int atoi(const char* s);

struct foo {
  int x;
  int y;
};

static int get_x(struct foo f) {
  f.x *= 2;
  return f.x;
}

static int get_y(struct foo f) {
  f.y *= 3;
  return f.y;
}

int main() {
  struct foo f = { atoi("1"), atoi("2") };
  int x = get_x(f);
  int y = get_y(f);
  printf("%d %d %d %d\n", x, y, f.x, f.y);
}
```
produces
```llvm
; ModuleID = '../test_struct_split.c'
source_filename = "../test_struct_split.c"
target datalayout = "e-m:e-p:32:32-i64:64-n32:64-S128"
target triple = "wasm32-unknown-wasi"

%struct.foo = type { i32, i32 }

@.str = private unnamed_addr constant [2 x i8] c"1\00", align 1
@.str.1 = private unnamed_addr constant [2 x i8] c"2\00", align 1
@.str.2 = private unnamed_addr constant [13 x i8] c"%d %d %d %d\0A\00", align 1
@llvm.used = appending global [1 x i8*] [i8* bitcast (i32 ()* @__main_void to i8*)], section "llvm.metadata"

@__main_void = alias i32 (), i32 ()* @main

; Function Attrs: noinline nounwind optnone
define hidden i32 @main() #0 {
entry:
  %f = alloca %struct.foo, align 4
  %x2 = alloca i32, align 4
  %y4 = alloca i32, align 4
  %x = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 0
  %call = call i32 @atoi(i8* getelementptr inbounds ([2 x i8], [2 x i8]* @.str, i32 0, i32 0))
  store i32 %call, i32* %x, align 4
  %y = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 1
  %call1 = call i32 @atoi(i8* getelementptr inbounds ([2 x i8], [2 x i8]* @.str.1, i32 0, i32 0))
  store i32 %call1, i32* %y, align 4
  %call3 = call i32 @get_x(%struct.foo* byval(%struct.foo) align 4 %f)
  store i32 %call3, i32* %x2, align 4
  %call5 = call i32 @get_y(%struct.foo* byval(%struct.foo) align 4 %f)
  store i32 %call5, i32* %y4, align 4
  %0 = load i32, i32* %x2, align 4
  %1 = load i32, i32* %y4, align 4
  %x6 = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 0
  %2 = load i32, i32* %x6, align 4
  %y7 = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 1
  %3 = load i32, i32* %y7, align 4
  %call8 = call i32 (i8*, ...) @printf(i8* getelementptr inbounds ([13 x i8], [13 x i8]* @.str.2, i32 0, i32 0), i32 %0, i32 %1, i32 %2, i32 %3)
  ret i32 0
}

declare i32 @atoi(i8*) #1

; Function Attrs: noinline nounwind optnone
define internal i32 @get_x(%struct.foo* byval(%struct.foo) align 4 %f) #0 {
entry:
  %x = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 0
  %0 = load i32, i32* %x, align 4
  %mul = mul nsw i32 %0, 2
  store i32 %mul, i32* %x, align 4
  %x1 = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 0
  %1 = load i32, i32* %x1, align 4
  ret i32 %1
}

; Function Attrs: noinline nounwind optnone
define internal i32 @get_y(%struct.foo* byval(%struct.foo) align 4 %f) #0 {
entry:
  %y = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 1
  %0 = load i32, i32* %y, align 4
  %mul = mul nsw i32 %0, 3
  store i32 %mul, i32* %y, align 4
  %y1 = getelementptr inbounds %struct.foo, %struct.foo* %f, i32 0, i32 1
  %1 = load i32, i32* %y1, align 4
  ret i32 %1
}

declare i32 @printf(i8*, ...) #1

attributes #0 = { noinline nounwind optnone "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="generic" "unsafe-fp-math"="false" "use-soft-float"="false" }
attributes #1 = { "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="generic" "unsafe-fp-math"="false" "use-soft-float"="false" }

!llvm.module.flags = !{!0}
!llvm.ident = !{!1}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{!"clang version 12.0.1 (https://github.com/microsoft/vcpkg.git 2a31089e777fc187f1cc05338250b8e1810cfb52)"}
```
whose expected output is `2 6 1 2`.
Decompiling it via Rellic produces
```c
unsigned char _str[2] = "1\000";
unsigned char _str_1[2] = "2\000";
unsigned char _str_2[13] = "%d %d %d %d\n\000";
unsigned int main();
unsigned int atoi(unsigned char *arg0);
struct struct_foo {
    unsigned int field0;
    unsigned int field1;
};
unsigned int get_x(struct struct_foo *f);
unsigned int get_y(struct struct_foo *f);
unsigned int printf(unsigned char *arg0, ...);
unsigned int main() {
    struct struct_foo var0;
    unsigned int var1;
    unsigned int var2;
    unsigned int val3;
    unsigned int val4;
    unsigned int val5;
    unsigned int val6;
    unsigned int val7;
    val3 = atoi(_str);
    var0.field0 = val3;
    val4 = atoi(_str_1);
    var0.field1 = val4;
    val5 = get_x(&var0);
    var1 = val5;
    val6 = get_y(&var0);
    var2 = val6;
    val7 = printf(_str_2, var1, var2, var0.field0, var0.field1);
    return 0U;
}
unsigned int get_x(struct struct_foo *f) {
    f->field0 = f->field0 * 2U;
    return f->field0;
}
unsigned int get_y(struct struct_foo *f) {
    f->field1 = f->field1 * 3U;
    return f->field1;
}
```
whose output (after a bit of massaging to correct the type inaccuracies) is `2 6 2 6`, which is incorrect as the structs are passed by reference instead of by value

frabert:
Fixes #195


pgoodman:
@frabert Are any of the things in tyat build_linux for ubuntu 20.04 on llvm-12 issues?

frabert:
The issue seems to be about dangling references or something like that...
```While deleting: i32* %SDVO_CMD_GET_MAX_OVERSCAN_H
Use still stuck around after Def is destroyed:  <badref> = load i32, i32* @SDVO_CMD_GET_MAX_OVERSCAN_H, align 4, !dbg !DILocation(line: 74, column: 8, scope: <0x4f579c0>), !tbaa <0x4f5a598>
rellic-decomp-12.0: /__w/cxx-common/cxx-common/vcpkg/buildtrees/llvm-12/src/org-12.0.1-5ce32a7629.clean/llvm/lib/IR/Value.cpp:100: llvm::Value::~Value(): Assertion `materialized_use_empty() && "Uses remain when a value is destroyed!"' failed.```

File: lib/AST/FixByVal.cpp [hunk 1]
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+  llvm::SmallVector<llvm::ReturnInst*> rets{};
+  llvm::CloneFunctionInto(new_func, &old_func, map, true, rets);
+  builder.CreateBr(new_func->getBasicBlockList().getNextNode(*bb));
+  subst[&old_func] = new_func;
+  old_func.removeFromParent();
+}
+
+void FixByVal::visitCallInst(llvm::CallInst& call) {
```
pgoodman, line 74: I think we need to handle /all/ uses of the function, and replace them, not just calls. This is much trickier than it seems, though, because functions can be used inside of constants. You can probably copy some of the constant cloning/replacement code from remill [1]. It ends up being quite a bit of code, but I think we want to make sure we handle situations like function pointers in code and data.

[1] https://github.com/lifting-bits/remill/blob/master/lib/BC/Util.cpp#L1576-L1616

pgoodman (changes requested):

lib/AST/FixByVal.cpp
pgoodman: It is probably more complicated than what my the cloning code does; you'll need to decide/choose between:
 - if there's a function pointer in a constant, then do we cast the new version of the function to the old type; or
 - do we re-type the constant/global, and then find the effects of that.

I'm realizing that an llvm transform is opening us up to some really tricky issues. Maybe one possibility is this, that side-steps our things:
 - the original is retained, but changed so that it does a load of the `byval` pointer and calls the new version.

pgoodman: Or, we fall back to modifying how `byval`-attributed parameters are handled in Rellic.

lib/AST/IRToASTVisitor.cpp
```diff
@@ -398,8 +398,28 @@ void IRToASTVisitor::VisitArgument(llvm::Argument &arg) {
...
+  auto argtype{arg.getType()};
+  if (arg.hasByValAttr()) {
+    argtype = llvm::cast<llvm::PointerType>(argtype)->getElementType();
+  }
   // Create a declaration
-  parm = ast.CreateParamDecl(fdecl, GetQualType(arg.getType()), name);
+  parm = ast.CreateParamDecl(fdecl, GetQualType(argtype), name);
+}
+
+static llvm::FunctionType *GetFixedFunctionType(llvm::Function &func) {
```
pgoodman, line 409: Add a comment with a bit of an example of what `byval` arguments look like, and how we re-interpret them as actual values.

lib/AST/IRToASTVisitor.cpp
```diff
@@ -509,6 +534,24 @@ void IRToASTVisitor::visitGetElementPtrInst(llvm::GetElementPtrInst &inst) {
   auto indexed_type{inst.getPointerOperandType()};
   auto base{GetOperandExpr(inst.getPointerOperand())};
 
+  if (auto arg = llvm::dyn_cast<llvm::Argument>(inst.getPointerOperand())) {
```
pgoodman, line 537: This seems to handle the GEP case, but what about `ptrtoint`, `bitcast`, tail-calls, etc? What about an alternative approach where we pretend that arguments are like `phi` nodes and thus need to be "solved" via `alloca`s. Then, the `byval` arguments become the normal case, and the non-byval ones are immediately put into `alloca`s. Or is this crazy? Mostly, I'm worried about losing track of the byval pointer somehow.

lib/AST/IRToASTVisitor.cpp
```diff
@@ -398,8 +398,28 @@ void IRToASTVisitor::VisitArgument(llvm::Argument &arg) {
   // Get parent function declaration
   auto func{arg.getParent()};
   auto fdecl{clang::cast<clang::FunctionDecl>(GetOrCreateDecl(func))};
+  auto argtype{arg.getType()};
+  if (arg.hasByValAttr()) {
+    argtype = llvm::cast<llvm::PointerType>(argtype)->getElementType();
```
pgoodman, line 403: Is the pointer element type communicated via the `byval` attribute when opaque pointer types are used?

lib/AST/IRToASTVisitor.cpp
pgoodman: Here's a new test case for you: https://godbolt.org/z/6fz8hbzd9

```c++
struct big_base {
  int x[32];
};
struct big_derived : public big_base {

};
extern double bar(struct big_base y);
double foo(struct big_derived x) {
  return bar(x);
}
```

Which produces the following:

```llvm
; Function Attrs: mustprogress uwtable
define dso_local double @_Z3foo11big_derived(%struct.big_derived* nocapture readonly byval(%struct.big_derived) align 8 %0) local_unnamed_addr #0 {
  %2 = getelementptr inbounds %struct.big_derived, %struct.big_derived* %0, i64 0, i32 0
  %3 = tail call double @_Z3bar8big_base(%struct.big_base* nonnull byval(%struct.big_base) align 8 %2)
  ret double %3
}
```

pgoodman (changes requested):

lib/AST/IRToASTVisitor.cpp
pgoodman: And another: https://godbolt.org/z/79E9Mc1dW

```c++
struct big_base {
  int x[32];
};
extern double bar(struct big_base y);
double foo(struct big_base x) {
  return bar(x);
}
```

This one has no GEP:
```llvm
; Function Attrs: mustprogress uwtable
define dso_local double @_Z3foo8big_base(%struct.big_base* nocapture readonly byval(%struct.big_base) align 8 %0) local_unnamed_addr #0 {
  %2 = tail call double @_Z3bar8big_base(%struct.big_base* nonnull byval(%struct.big_base) align 8 %0)
  ret double %2
}
```

lib/AST/IRToASTVisitor.cpp
frabert: I'm not sure what you mean by 'opaque pointer types'
As the call is by value, the type must be complete, I'm not sure how the pointer type could become opaque

lib/AST/IRToASTVisitor.cpp
frabert: I've switched to using a more 'general' approach in generating the auxiliary pointer value, this should be robust enough to work for other instructions as well.

I've added a specific test type that decompiles bitcode directly, another approach could be to instead test roundtripping with full optimization too instead.

lib/AST/FixByVal.cpp
frabert: I guess the latter was the easiest one :)

lib/AST/IRToASTVisitor.cpp
pgoodman: I think newer LLVM versions support pointer types without an element type. Here's a concrete `byval`-related example: https://llvm.org/docs/OpaquePointers.html#i-still-need-pointee-types

frabert: Addressed in latest commit

pgoodman (approved):

PR #212, opened 2021-12-29
Decompiler API

frabert:


include/rellic/Decompiler.h
```diff
@@ -0,0 +1,41 @@
...
+};
+
+struct DecompilationResult {
+  std::unique_ptr<llvm::Module> module;
+  std::unique_ptr<clang::ASTUnit> ast;
+  StmtToIRMap stmt_provenance_map;
+};
+
+struct DecompilationError {
+  std::unique_ptr<clang::ASTUnit> ast;
```
pgoodman, line 35: Also give back the module?

tools/decomp/Decomp.cpp
```diff
@@ -6,37 +6,18 @@
...
 #include <llvm/Support/raw_ostream.h>
-#include <llvm/Transforms/Utils.h>
-#include <llvm/Transforms/Utils/Local.h>
 
-#include <memory>
-#include <sstream>
+#include <iostream>
 #include <system_error>
 
 #include "rellic/AST/ASTPrinter.h"
```
pgoodman, line 18: Lets get rid of the AST printer from rellic.

pgoodman: And just use normal pretty printing.

pgoodman (changes requested):

frabert:
I've converted `LOG(FATAL)` and `CHECK()` into throw statements more or less arbitrary, judging by whether the checks originally looked like internal invariant checking or unsupported features checking

lib/AST/ASTBuilder.cpp
```diff
@@ -178,7 +183,7 @@ clang::StringLiteral *ASTBuilder::CreateStrLit(std::string val) {
 clang::FloatingLiteral *ASTBuilder::CreateFPLit(llvm::APFloat val) {
   auto size{llvm::APFloat::getSizeInBits(val.getSemantics())};
   auto type{GetLeastRealTypeForBitWidth(size)};
-  CHECK(!type.isNull()) << "Unable to infer type for given value.";
+  { CHECK_THROW(!type.isNull()) << "Unable to infer type for given value."; }
```
pgoodman, line 186: I think if we wanted to avoid the `{` then you could maybe do something like:

```
template <typename E>
class ThrowerStream {
  std::stringstream &ss;
  bool can_throw{true};
  inline explicit ThrowerStream(std::stringstream &ss_=std::stringstream())
    : ss(ss_) {}

  ~ThrowerStream(void) {
    if (can_throw) {
      throw E(ss.str());
    }
  }

  template <typename T>
  ThrowerStream operator<<(T &&val) {
    can_throw = false;
    ss << std::forward<T>(val);
    return ThrowerStream(ss);
  }
};
```

pgoodman (changes requested):

lib/AST/ASTBuilder.cpp
frabert: I'm loving these tricks :)

frabert:
@pgoodman Can you report here if there are more `LOG(FATAL)`/`CHECK(...)` getting triggered when using rellic in IRENE? I feel I don't have a good way to test for that right now

include/rellic/Decompiler.h
```diff
@@ -17,18 +17,18 @@
...
-using IRToDeclMap = std::unordered_map<llvm::Value*, clang::ValueDecl*>;
-
 struct DecompilationOptions {
   bool lower_switches = false;
   bool remove_phi_nodes = false;
   bool disable_z3 = false;
 };
 
 struct DecompilationResult {
+  using StmtToIRMap = std::unordered_map<clang::Stmt*, const llvm::Value*>;
```
pgoodman, line 27: Make these `const clang::` as well.

pgoodman (changes requested):

pgoodman (approved):

frabert:
When/if the automated checks pass, I am going to call this PR ready for review and merge, and end the scope creep

pgoodman (approved):

PR #224, opened 2022-02-14
Fix issue with `LoopRefine`

frabert:
`LoopRefine` recognizes patterns like
```c
while(1) {
  if(cond) {
    break;
  }
  body;
}
```
and turns it into
```c
while(!cond) {
  body;
}
```

Unfortunately it also discards the `else` branch of the `if` statement used for the loop refinements, which leads to incorrect results when multiple `if` statements have been merged into one by a previous pass like `ConditionBasedRefine`. This fixes that.

### DIFF
--- a/lib/AST/LoopRefine.cpp
+++ b/lib/AST/LoopRefine.cpp
@@ -11,6 +11,9 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 
+#include <algorithm>
+#include <iterator>
+
 #include "rellic/AST/ASTBuilder.h"
 #include "rellic/AST/InferenceRule.h"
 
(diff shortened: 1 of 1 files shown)


lib/AST/LoopRefine.cpp
```diff
@@ -51,9 +54,14 @@ class WhileRule : public InferenceRule {
         << "Substituted WhileStmt is not the matched WhileStmt!";
 
     auto comp{clang::cast<clang::CompoundStmt>(loop->getBody())};
-    auto cond{clang::cast<clang::IfStmt>(comp->body_front())->getCond()};
-    std::vector<clang::Stmt *> new_body(comp->body_begin() + 1,
-                                        comp->body_end());
+    auto ifstmt{clang::cast<clang::IfStmt>(comp->body_front())};
+    auto cond{ifstmt->getCond()};
+    std::vector<clang::Stmt *> new_body;
+    if (ifstmt->getElse()) {
```
surovic, line 60: Just a nitpick:
```C
if(auto else_stmt = ifstmt->getElse()) {
  new_body.push_back(else_stmt);
}
```

lib/AST/LoopRefine.cpp
```diff
@@ -85,9 +93,13 @@ class DoWhileRule : public InferenceRule {
         << "Substituted WhileStmt is not the matched WhileStmt!";
 
     auto comp{clang::cast<clang::CompoundStmt>(loop->getBody())};
-    auto cond{clang::cast<clang::IfStmt>(comp->body_back())->getCond()};
+    auto ifstmt{clang::cast<clang::IfStmt>(comp->body_back())};
+    auto cond{ifstmt->getCond()};
     std::vector<clang::Stmt *> new_body(comp->body_begin(),
                                         comp->body_end() - 1);
+    if (ifstmt->getElse()) {
```
surovic, line 100: Same nitpick as above:
```C
if(auto else_stmt = ifstmt->getElse()) {
  new_body.push_back(else_stmt);
}
```

lib/AST/LoopRefine.cpp
```diff
@@ -129,6 +141,9 @@ class NestedDoWhileRule : public InferenceRule {
 
     std::vector<clang::Stmt *> do_body(comp->body_begin(),
                                        comp->body_end() - 1);
+    if (cond->getElse()) {
```
surovic, line 144: Same as above.

surovic (approved):

lib/AST/LoopRefine.cpp
frabert: Done!

Issue #225, opened 2022-02-14
Could `NestedCondProp` also propagate `while` conditions?


frabert:
I'm experimenting locally with a version of `NestedCondProp` which propagates conditions for `while` statements too, instead of just `if`s. It seems to work ok, as in all the tests pass and the output improves, but I'm not sure if it is ok in general.

Also, converting the `parent_conds[child] = ast.CreateLNot(cond)` with something like

```c++
static clang::Expr *Negate(rellic::ASTBuilder &ast, clang::Expr *expr) {
  if (auto binop = clang::dyn_cast<clang::BinaryOperator>(expr)) {
    auto opc = clang::BinaryOperator::negateComparisonOp(binop->getOpcode());
    return ast.CreateBinaryOp(opc, binop->getLHS(), binop->getRHS());
  } else if (auto unop = clang::dyn_cast<clang::UnaryOperator>(expr)) {
    if (unop->getOpcode() == clang::UO_LNot) {
      return unop->getSubExpr();
    }
  }
  return ast.CreateLNot(expr);
}

/* ... */
/* ... */
parent_conds[child] = Negate(ast, child);
```
creates more opportunities for the replacement to find matches and improve conditions.

PR #227, opened 2022-02-15
`NestedCondProp`: propagate `while` conditions too

frabert:
As described in #225 this PR modifies `NestedCondProp` so that `while` conditions are propagated, in addition to `if` conditions.


pgoodman:
Seems to be some Linux failures?

frabert:
There were some performance issues with the Z3 simplification pass, so I added some caching. Also, `if(0U) {...}` statements were causing divergence when going through `ConditionBasedRefinement`, so I pruned them as they are dead code anyway.

PR #231, opened 2022-02-18
Update to LLVM 13

frabert:
The module printing code in `rellic-xref` was flaky and got replaced by one which should be forward-compatible but unfortunately is not as nice (no syntax highlighting).
It would be nice to have a more robust code for AST printing as well.


frabert:
@ekilmer I think `clang-13` is only available on Ubuntu 21.10+, do you think we could hardcode `clang-12` for anghabench? Or perhaps we could add the official LLVM repo to APT and use that

README.md
```diff
@@ -28,8 +28,8 @@ Most of Rellic's dependencies can be provided by the [cxx-common](https://github
 | [CMake](https://cmake.org/) | 3.14+ |
```
ekilmer, line 28: While we're in here...
```suggestion
| [CMake](https://cmake.org/) | 3.21+ |
```

README.md
```diff
@@ -62,7 +62,7 @@ sudo apt install \
      doctest-dev
 ```
 
-If the distribution you're on doesn't include a recent release of CMake (3.14 or later), you'll need to install it. For Ubuntu, see here https://apt.kitware.com/.
+If the distribution you're on doesn't include a recent release of CMake (3.14 or later), you'll need to install it. For Ubuntu, see here <https://apt.kitware.com/>.
```
ekilmer, line 65:
```suggestion
If the distribution you're on doesn't include a recent release of CMake (3.21 or later), you'll need to install it. For Ubuntu, see here <https://apt.kitware.com/>.
```

ekilmer:
> @ekilmer I think `clang-13` is only available on Ubuntu 21.10+, do you think we could hardcode `clang-12` for anghabench? Or perhaps we could add the official LLVM repo to APT and use that

I think adding the apt repo should be fine. I'm not super familiar with the Anghabench testing with regards to upgrading versions... but looks like CI is working passing now?

Ping @artemdinaburg for anghabench things

PR #234, opened 2022-02-23
Fix iterator invalidation

frabert:
Thanks @pgoodman!

### DIFF
--- a/lib/AST/Util.cpp
+++ b/lib/AST/Util.cpp
@@ -68,8 +68,10 @@ bool IsEquivalent(clang::ASTContext &ctx, clang::Stmt *a, clang::Stmt *b) {
 
 void CopyProvenance(clang::Stmt *from, clang::Stmt *to, StmtToIRMap &map) {
   auto range{map.equal_range(from)};
-  std::vector<std::pair<clang::Stmt *, llvm::Value *>> pairs{range.first,
-                                                             range.second};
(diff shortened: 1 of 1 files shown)


pgoodman (approved):

lib/AST/Util.cpp
```diff
@@ -67,9 +67,14 @@ bool IsEquivalent(clang::ASTContext &ctx, clang::Stmt *a, clang::Stmt *b) {
 }
 
 void CopyProvenance(clang::Stmt *from, clang::Stmt *to, StmtToIRMap &map) {
+  std::vector<std::pair<clang::Stmt *, llvm::Value *>> pairs;
   auto range{map.equal_range(from)};
   for (auto it{range.first}; it != range.second && it != map.end(); ++it) {
```
pgoodman, line 72: `pairs.insert(pairs.end(), range.first, range.second);`?

pgoodman (changes requested):

lib/AST/Util.cpp
frabert: I'm not sure that would work, I've found that I both had to check `range.second` _and_ `map.end()`, otherwise it'd sometimes run out of bounds

pgoodman: 👍

pgoodman (approved):

pgoodman (approved):

PR #241, opened 2022-03-03
Don't reuse Clang nodes

frabert:
Reusing AST nodes during code generation saves time and memory, but may hinder some transformation techniques in the future (because e.g. we can't mutate nodes with the assumption that such changes will only be local).

If anyone can think of a way to make `GenerateAST::GetOrCreateReachingCond` work without having to clone, it would be great.


lib/AST/Util.cpp
```diff
@@ -75,4 +78,141 @@ void CopyProvenance(clang::Stmt *from, clang::Stmt *to, StmtToIRMap &map) {
...
+  }
+
+  clang::Expr *VisitCompoundLiteralExpr(clang::CompoundLiteralExpr *expr) {
+    LOG(INFO) << "VisitCompoundLiteralExpr";
+    return ast.CreateCompoundLit(expr->getType(),
+                                 Visit(expr->getInitializer()));
+  }
+
+  clang::Expr *VisitParenExpr(clang::ParenExpr *expr) {
+    LOG(INFO) << "VisitParenExpr";
```
pgoodman, line 183: Make these `LOG(INFO)` into `DLOG(INFO)`, ideally removing them eventually because I don't see why we want logging in here.

pgoodman (changes requested):

pgoodman (approved):

PR #244, opened 2022-03-20
Create variables for values with multiple uses

frabert:


lib/AST/IRToASTVisitor.cpp
```diff
@@ -818,59 +840,75 @@ clang::Stmt *IRToASTVisitor::visitBinaryOperator(llvm::BinaryOperator &inst) {
...
     case llvm::BinaryOperator::Or:
-      return inst.getType()->isIntegerTy(1U) ? ast.CreateLOr(lhs, rhs)
-                                             : ast.CreateOr(lhs, rhs);
+      res = inst.getType()->isIntegerTy(1U) ? ast.CreateLOr(lhs, rhs)
+                                            : ast.CreateOr(lhs, rhs);
+      break;
 
     case llvm::BinaryOperator::Xor:
-      return ast.CreateXor(lhs, rhs);
+      res = ast.CreateXor(lhs, rhs);
```
pgoodman, line 869: Perhaps here we should check for XORing against an `llvm::ConstantInt`. If the constant is `ci->isMInusOne()` then emit logical `!` if the size is one, and bitwise `~` is the size isn't one.

pgoodman (approved):
Looks good. Added a possible thing we could do to XOR in a future PR, if the `ast.CreateXor` doesn't already do it.

lib/AST/IRToASTVisitor.cpp
frabert: That's probably something that could be done in a refinement pass

PR #246, opened 2022-03-23
Provide provenance info for `llvm::Use`s

frabert:


frabert:
Needs #247 to pass AnghaBench

lib/AST/IRToASTVisitor.cpp
```diff
@@ -691,29 +686,19 @@ clang::Stmt *IRToASTVisitor::visitGetElementPtrInst(
...
-  return ast.CreateNullStmt();
-}
-
-clang::Stmt *IRToASTVisitor::visitExtractValueInst(
+clang::Expr *IRToASTVisitor::visitExtractValueInst(
     llvm::ExtractValueInst &inst) {
   DLOG(INFO) << "visitExtractValueInst: " << LLVMThingToString(&inst);
 
-  auto base{GetOperandExpr(inst.getAggregateOperand())};
+  auto base{GetOperandExpr(inst.getOperandUse(0))};
```
surovic, line 701: Is there some other way to get the `llvm::Use` that would be more easily understandable? The indexes are kind of opaque to me personally.

lib/AST/GenerateAST.cpp
```diff
@@ -138,11 +139,11 @@ clang::Expr *GenerateAST::CreateEdgeCond(llvm::BasicBlock *from,
       auto br = llvm::cast<llvm::BranchInst>(term);
       if (br->isConditional()) {
         // Get the edge condition
-        result = ast_gen.GetOperandExpr(br->getCondition());
+        result = ast_gen.GetOperandExpr(*(br->op_end() - 3));
```
surovic, line 142: Is there some other way to get the llvm::Use that would be more easily understandable? The indexes are kind of opaque to me personally.

lib/AST/GenerateAST.cpp
```diff
@@ -203,12 +204,114 @@ clang::Expr *GenerateAST::GetOrCreateReachingCond(llvm::BasicBlock *block) {
   return cond;
 }
 
+class BlockVisitor : public llvm::InstVisitor<BlockVisitor, clang::Stmt *> {
```
surovic, line 207: What's the exact purpose of this class? How does it relate to `IRToASTVisitor` and how is it different?

surovic (changes requested):
Generally looks okay and I get what you're trying to do with the "SSAification" of the AST.

What I have qualms about is the mingling of "IR-to-C translation code" and "structurization code". `IRToASTVisitor.(h|cpp)` was meant to be the place where the IR translation code lived. `GenerateAST.(h|cpp)` was for structurization. The `BlockVisitor` class I think breaks this distinction. Personally, I would like for this kind of distinction to exist. One reason, because it makes it easy to reference the No More Gotos paper back and forth. Another reason is purely for conceptual partitioning of the code.

The distinctions don't have to be "translation" and "structurization". If you can come up and describe (optionally document) the conceptual distinctions and they make sense, anything is fine by me.

lib/AST/IRToASTVisitor.cpp
frabert: Not in general case I'm afraid... Some instructions expose a method to get the index, some of them only as `protected`, others not at all

lib/AST/GenerateAST.cpp
```diff
@@ -203,12 +204,114 @@ clang::Expr *GenerateAST::GetOrCreateReachingCond(llvm::BasicBlock *block) {
...
+
 StmtVec GenerateAST::CreateBasicBlockStmts(llvm::BasicBlock *block) {
   StmtVec result;
+  BlockVisitor visitor(*ast_ctx, ast, ast_gen, result, GetStmtToIRMap(),
+                       GetIRToValDeclMap());
   for (auto &inst : *block) {
-    if (auto stmt = ast_gen.GetOrCreateStmt(&inst)) {
-      result.push_back(stmt);
-    }
+    visitor.VisitInstruction(inst);
```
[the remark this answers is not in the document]
frabert: Done

include/rellic/AST/Util.h
```diff
@@ -45,9 +46,22 @@ size_t GetNumDecls(clang::DeclContext *decl_ctx) {
 }
 
 using StmtToIRMap = std::unordered_multimap<clang::Stmt *, llvm::Value *>;
-void CopyProvenance(clang::Stmt *from, clang::Stmt *to, StmtToIRMap &map);
+using ExprToUseMap = std::unordered_multimap<clang::Expr *, llvm::Use *>;
```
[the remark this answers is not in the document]
frabert: Done

lib/AST/IRToASTVisitor.cpp
```diff
@@ -246,88 +263,82 @@ clang::Expr *IRToASTVisitor::CreateLiteralExpr(llvm::Constant *constant) {
...
-        temp = ast.CreateVarDecl(fdecl, GetQualType(arg->getType()),
-                                 argdecl->getName().str() + "_ptr");
-        temp->setInit(addr_of_arg);
-        fdecl->addDecl(temp);
-      }
 
-      return ast.CreateDeclRef(temp);
-    } else {
-      return ref;
+  auto Wrapper{[&]() -> clang::Expr * {
```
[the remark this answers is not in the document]
frabert: Done!

pgoodman (approved):

include/rellic/AST/IRToASTVisitor.h
```diff
@@ -26,70 +26,49 @@
...
+  IRToASTVisitor(clang::ASTUnit &unit, Provenance &provenance);
 
-  clang::Stmt *GetOrCreateStmt(llvm::Value *val);
   clang::Decl *GetOrCreateDecl(llvm::Value *val);
-  clang::Expr *GetOperandExpr(llvm::Value *val);
-
-  StmtToIRMap &GetStmtToIRMap() { return provenance; }
-  IRToValDeclMap &GetIRToValDeclMap() { return value_decls; }
-  IRToTypeDeclMap &GetIRToTypeDeclMap() { return type_decls; }
+  clang::Expr *GetOperandExpr(llvm::Use &val);
```
surovic, line 50: Probably should be renamed to `CreateOperandExpr(llvm::Use &val)` since it no longer "gets" anything from a map.

lib/AST/IRToASTVisitor.cpp
```diff
@@ -505,20 +459,57 @@ void IRToASTVisitor::VisitFunctionDecl(llvm::Function &func) {
...
   for (auto &arg : func.args()) {
     auto parm{clang::cast<clang::ParmVarDecl>(GetOrCreateDecl(&arg))};
     params.push_back(parm);
   }
 
-  decl->getAsFunction()->setParams(params);
+  auto fdecl{decl->getAsFunction()};
+  fdecl->setParams(params);
+
+  for (auto &inst : llvm::instructions(func)) {
```
surovic, line 471: Can we do this in `BlockVisitor` as we create instructions? I'd prefer `VisitFunctionDecl` to not create AST nodes that relate to the function body.

surovic (changes requested):
As much as I don't like to bloat up `IRToASTVisitor.(h|cpp)` I think it makes most sense to put `BlockVisitor` there. Maybe even move `clang::Expr *visitXYZ` methods into a separate class, named something like `ExprGen`. Move `clang::Stmt *visitXYZ` methods from `BlockVisitor` into `StmtGen`. And finally make an `IRToASTVisitor` class use both. The logic from `BlockVisitor::visitBasicBlock` could be put into a `StmtVec IRToASTVisitor::CreateBasicBlockStmts(llvm::BasicBlock&)` or similar. What do you think?

lib/AST/IRToASTVisitor.cpp
frabert: Things are easier if we can create all declarations at once, and to me `VisitFunctionDecl` looks like the best place to do so right now...

lib/AST/IRToASTVisitor.cpp
```diff
@@ -24,392 +27,885 @@
...
+
+  Provenance &provenance;
+  size_t num_literal_structs = 0;
+  size_t num_declared_structs = 0;
+
+ public:
+  ExprGen(clang::ASTUnit &unit, Provenance &provenance)
+      : ast_ctx(unit.getASTContext()), ast(unit), provenance(provenance) {}
+
+  clang::QualType GetQualType(llvm::Type *type) {
```
surovic, line 44: Can we factor out method bodies like `GetQualType` out of the class declaration? Again, no other but for consistency and/or neatness's sake. I promise this is the last piece of annoyance :eyes:

surovic (approved):

lib/AST/IRToASTVisitor.cpp
frabert: Do you mean making them `static` or separating all the declarations of `ExprGen` and `StmtGen` from their definitions?

surovic:
> Had to put `VisitGlobalVar` inside `ExprGen` so that when recursively visiting the initializers for global variables we "automagically" get topologically sorted declarations

Neat! Think we can merge?

frabert:
Yep, tests are passing!

PR #248, opened 2022-03-24
Add CMakePresets support for building with ASAN

artemdinaburg:
* Support for building against an ASAN cxx-common


frabert:
Would it make sense to have a CI asan run?

pgoodman (approved):

PR #254, opened 2022-04-05
Preprocess reaching conditions

frabert:
Ran into a bug while trying to roundtrip optimized builds of the tests.
Preprocesses the reaching conditions for blocks in RPO order to fix it.
Thanks @surovic for the collab session!


scripts/roundtrip.py
```diff
@@ -117,7 +117,7 @@ class TestRoundtrip(unittest.TestCase):
 
     def test_generator(path):
         def test(self):
-            roundtrip(self, args.rellic, path, args.clang, args.timeout, args.translate_only)
+            roundtrip(self, args.rellic, path, args.clang, args.timeout, args.translate_only, ["-O3"])
```
surovic, line 120: Maybe you could try running `-O1` and `-O2` as well? See what it does.

pgoodman (approved):

frabert:
So I have a very quick fix for the current failures in AnghaBench, which is to download more RAM. Unfortunately it's not really that realistic so I'll have to find a way to prune the reaching conditions as soon as they are generated.

surovic:
> So I have a very quick fix for the current failures in AnghaBench, which is to download more RAM. Unfortunately it's not really that realistic so I'll have to find a way to prune the reaching conditions as soon as they are generated.

Is this caused by the changes in this PR?

frabert:
Looks like so. Separating `GetReachingCondition` from `CreateReachIngCondition` might help, but we're still running on the edge of timeouts, even in the absence of OOM kills

scripts/roundtrip.py
frabert: Added tests for all optimization levels

lib/AST/GenerateAST.cpp
```diff
@@ -502,6 +589,7 @@ GenerateAST::Result GenerateAST::run(llvm::Function &func,
   // structurization
   llvm::ReversePostOrderTraversal<llvm::Function *> rpo(&func);
   rpo_walk.assign(rpo.begin(), rpo.end());
+  GetOrCreateReachingCond(rpo_walk[0]);
```
frabert, line 592: Walking all the blocks in `rpo_walk` causes the condition explosion. Only inspecting the first block seems to work, but I cannot say I'm sure it's correct in general.

frabert:
I'm officially at a loss on how to reproduce the issue locally. I have literally uploaded the macOS artifacts and tested them locally and they decompile the right way 😖

pgoodman (approved):

PR #255, opened 2022-04-06
Directly handle `switch` instructions

frabert:


pgoodman (approved):

lib/AST/ASTBuilder.cpp
```diff
@@ -536,4 +536,21 @@ clang::NullStmt *ASTBuilder::CreateNullStmt() {
   return new (ctx) clang::NullStmt(clang::SourceLocation());
 }
 
+clang::SwitchStmt *ASTBuilder::CreateSwitchStmt(clang::Expr *cond) {
```
surovic, line 539: Are there no methods in `clang::Sema` to create these? Just to get some extra bit of semantic checking.

lib/AST/GenerateAST.cpp
```diff
@@ -148,6 +148,33 @@ clang::Expr *GenerateAST::CreateEdgeCond(llvm::BasicBlock *from,
         }
       }
     } break;
+    // Switches
+    case llvm::Instruction::Switch: {
+      auto sw{llvm::cast<llvm::SwitchInst>(term)};
+      clang::Expr *edge_cond{nullptr};
+      for (auto sw_case : sw->cases()) {
+        auto cond{ast_gen.CreateOperandExpr(sw->getOperandUse(0))};
+        clang::Expr *sub;
```
surovic, line 157: Initialize pls :)

surovic (changes requested):
Can you describe what the idea behind structuring "switch regions" is? I'm asking because my original idea wrt to switch statements and instructions was to lower switch instructions into branches, generate reaching conditions and then create switch statements using "condition aware refinement" from the no more gotos paper. I do understand that switch lowering messes with IR<>C provenance.

lib/AST/ASTBuilder.cpp
frabert: `clang::Sema::CheckSwitchCondition` works fine, but `clang::Sema::ActOnCaseExpr` only works on `case`s that are already in a switch statement, I believe

surovic: Yeah, that's not surprising. Sometimes we can use sema, sometimes we can't. Hmm. Would it be practical to create the whole `switch` in one go with all the cases? So that we'd have something like `CreateSwitchStmt(clang::Expt *cond, std::vector<std::pair<clang::Expr*, clang::Stmt*>> &cases)` where the pair contains the case ordinal and case code.

frabert: Hrm it's not that easy, the statements under a `case` are not actually substatements of the `case` itself, the structure is more flat, with all the cases and their statements belonging to the compound statement of the switch itself

frabert:
I can't really find examples to break this mechanism, but if anyone can please report!

surovic (approved):

surovic:
LGTM! We should look into condition aware refinement from no more gotos though, since that might offer a more general approach to generating switch statements.

PR #259, opened 2022-04-19
General maintenance patches

frabert:


lib/AST/IRToASTVisitor.cpp
```diff
@@ -1155,11 +1155,22 @@ void IRToASTVisitor::VisitFunctionDecl(llvm::Function &func) {
       var = ast.CreateVarDecl(
           fdecl, expr_gen.GetQualType(alloca->getAllocatedType()), name);
       fdecl->addDecl(var);
-    } else if (inst.hasNUsesOrMore(2) || llvm::isa<llvm::CallInst>(inst) ||
+    } else if ((llvm::isa<llvm::CallInst>(inst) && inst.hasNUsesOrMore(2)) ||
```
frabert, line 1158: @surovic I think that until we are fully on board with the SSAification of the source code and we have a way to delete the unnecessary temporary variables, we are better off only generating the strictly necessary ones. In this way, we still have the correctness guarantees while keeping source code readable: I don't think that putting reused `add`s or things like that into variables helped much.

frabert: Upon further inspection, it _is_ necessary to keep things as they are 🙃

surovic (approved):

PR #262, opened 2022-04-23
Fixes #261

### LINKED ISSUE #261: Probable bug in `LoopRefine`
I suspect `LoopRefine` contains a bug. I cannot give a specific repro case, but when trying to improve condition propagation I ran into an issue with loops with this form:
```
while(1U) {
  a = sth;
  if(a == 10) {
    break;
  }
  if(a != 10) {
    body;
  }
}
```
getting turned into
```
do {
  a = sth;
  body;
} while(a != 10);
```
which is incorrect because the guard is not checked after `a = sth;`.

frabert:
Also improves the amount of propagated conditions by `NestedCondProp`

### DIFF
--- a/lib/AST/LoopRefine.cpp
+++ b/lib/AST/LoopRefine.cpp
@@ -100,10 +100,12 @@ class DoWhileRule : public InferenceRule {
     auto cond{ifstmt->getCond()};
     std::vector<clang::Stmt *> new_body(comp->body_begin(),
                                         comp->body_end() - 1);
-    if (auto else_stmt = ifstmt->getElse()) {
-      new_body.push_back(else_stmt);
-    }
     ASTBuilder ast(unit);
+    if (auto else_stmt = ifstmt->getElse()) {
+      auto cond_inv{ast.CreateLNot(cond)};
+      CopyProvenance(cond, cond_inv, provenance.use_provenance);
+      new_body.push_back(ast.CreateIf(cond_inv, else_stmt));
+    }
     auto cond_inv{ast.CreateLNot(cond)};
     CopyProvenance(cond, cond_inv, provenance.use_provenance);
     return ast.CreateDo(cond_inv, ast.CreateCompoundStmt(new_body));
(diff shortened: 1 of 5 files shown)


pgoodman (approved):

lib/AST/Util.cpp
```diff
@@ -316,6 +316,26 @@ clang::Expr *Clone(clang::ASTUnit &unit, clang::Expr *expr,
   return CHECK_NOTNULL(cloner.Visit(CHECK_NOTNULL(expr)));
 }
 
+static clang::Expr *StripParens(clang::Expr *expr) {
```
surovic, line 319: Can this be replaced by `clang::Expr::IgnoreParens()` ?

lib/AST/Util.cpp
```diff
@@ -316,6 +316,26 @@ clang::Expr *Clone(clang::ASTUnit &unit, clang::Expr *expr,
...
+static clang::Expr *ApplyLNot(rellic::ASTBuilder &ast, clang::Expr *expr) {
+  if (auto unop = clang::dyn_cast<clang::UnaryOperator>(expr)) {
+    if (unop->getOpcode() == clang::UO_LNot) {
+      return unop->getSubExpr();
+    }
+  }
+  return ast.CreateLNot(expr);
+}
+
+clang::Expr *Negate(rellic::ASTBuilder &ast, clang::Expr *expr) {
```
surovic, line 335: Do we really need to do this outside of `ExprCombine` ? There should be rewriting rules for this already, right?

lib/AST/LoopRefine.cpp
```diff
@@ -100,10 +100,12 @@ class DoWhileRule : public InferenceRule {
...
                                         comp->body_end() - 1);
+    ASTBuilder ast(unit);
     if (auto else_stmt = ifstmt->getElse()) {
-      new_body.push_back(else_stmt);
+      auto cond_inv{ast.CreateLNot(cond)};
+      CopyProvenance(cond, cond_inv, provenance.use_provenance);
+      new_body.push_back(ast.CreateIf(cond_inv, else_stmt));
     }
-    ASTBuilder ast(unit);
     auto cond_inv{ast.CreateLNot(cond)};
```
surovic, line 109: Am I right that we have `auto cond_inv{ast.CreateLNot(cond)}` twice here because of provenance tracking?

lib/AST/LoopRefine.cpp
```diff
@@ -147,11 +149,13 @@ class NestedDoWhileRule : public InferenceRule {
...
+    ASTBuilder ast(unit);
     if (auto else_stmt = cond->getElse()) {
-      do_body.push_back(else_stmt);
+      auto cond_inv{ast.CreateLNot(cond->getCond())};
+      CopyProvenance(cond->getCond(), cond_inv, provenance.use_provenance);
+      do_body.push_back(ast.CreateIf(cond_inv, else_stmt));
     }
 
-    ASTBuilder ast(unit);
     auto do_cond{ast.CreateLNot(cond->getCond())};
```
surovic, line 159: Same as `cond_inv` above.

surovic (changes requested):

lib/AST/Util.cpp
frabert: Since we changed the replacement mechanism in `NCP` to be syntactical, in this situation
```
if(!(a == b)) {
  ...
} else {
  if(a == b) {
    ...
  }
}
```
the condition would not be propagated to the `else` branch. The conditions need to be stripped down to the "bare minimum"

lib/AST/Util.cpp
frabert: Done!

lib/AST/Util.cpp
surovic: Got it!

surovic (approved):

PR #263, opened 2022-04-24
Fix bug in reaching conditions generation

frabert:
As suspected, computing reaching conditions via fixpoint is indeed necessary, but was not performed previously due to performance concerns. I've finally been able to find the right Z3 incantation to make conditions simplify enough to compute them sufficiently fast (on my machine at least).

### DIFF
--- a/include/rellic/AST/GenerateAST.h
+++ b/include/rellic/AST/GenerateAST.h
@@ -47,6 +47,7 @@ class GenerateAST : public llvm::AnalysisInfoMixin<GenerateAST> {
 
   std::map<BBEdge, unsigned> z_edges;
   std::unordered_map<llvm::BasicBlock *, unsigned> reaching_conds;
+  bool reaching_conds_changed{true};
   std::unordered_map<llvm::BasicBlock *, clang::IfStmt *> block_stmts;
   std::unordered_map<llvm::Region *, clang::CompoundStmt *> region_stmts;
 
@@ -61,7 +62,8 @@ class GenerateAST : public llvm::AnalysisInfoMixin<GenerateAST> {
                                     llvm::ConstantInt *c);
 
   z3::expr GetOrCreateEdgeCond(llvm::BasicBlock *from, llvm::BasicBlock *to);
-  z3::expr GetOrCreateReachingCond(llvm::BasicBlock *block);
+  z3::expr GetReachingCond(llvm::BasicBlock *block);
+  void CreateReachingCond(llvm::BasicBlock *block);
 
   clang::Expr *ConvertExpr(z3::expr expr);
 
--- a/lib/AST/GenerateAST.cpp
+++ b/lib/AST/GenerateAST.cpp
@@ -245,16 +245,8 @@ z3::expr GenerateAST::GetReachingCond(llvm::BasicBlock *block) {
 }
 
 void GenerateAST::CreateReachingCond(llvm::BasicBlock *block) {
-  auto Prove{[this](z3::expr expr) {
-    z3::goal goal(*z_ctx);
-    goal.add((!expr).simplify());
-    auto app{z3::tactic(*z_ctx, "sat").apply(goal)};
-    CHECK(app.size() == 1) << "Unexpected multiple goals in application!";
-    return app[0].is_decided_unsat();
-  }};
-
-  auto Simplify{[this, &Prove](z3::expr expr) {
-    if (Prove(expr)) {
+  auto Simplify{[this](z3::expr expr) {
+    if (Prove(*z_ctx, expr)) {
       return z_ctx->bool_val(true);
     }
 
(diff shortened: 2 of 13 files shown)


lib/AST/GenerateAST.cpp
```diff
@@ -236,36 +236,68 @@ z3::expr GenerateAST::GetOrCreateEdgeCond(llvm::BasicBlock *from,
...
 
-      reaching_conds[block] = z_exprs.size();
-      z_exprs.push_back(cond.simplify());
-    } else {
-      auto cond{z_ctx->bool_val(true)};
+  return z_exprs[reaching_conds[block]];
+}
+
+void GenerateAST::CreateReachingCond(llvm::BasicBlock *block) {
+  auto Prove{[this](z3::expr expr) {
```
surovic, line 248: I have a feeling that we use this piece of code at several places. Is this correct? Can we factor it out into `Utils.cpp` ?

lib/AST/GenerateAST.cpp
```diff
@@ -236,36 +236,68 @@ z3::expr GenerateAST::GetOrCreateEdgeCond(llvm::BasicBlock *from,
...
+    z3::goal goal(*z_ctx);
+    goal.add((!expr).simplify());
+    auto app{z3::tactic(*z_ctx, "sat").apply(goal)};
+    CHECK(app.size() == 1) << "Unexpected multiple goals in application!";
+    return app[0].is_decided_unsat();
+  }};
+
+  auto Simplify{[this, &Prove](z3::expr expr) {
```
surovic, line 256: Less sure of this one, but same as above.

lib/AST/GenerateAST.cpp
```diff
@@ -590,7 +616,12 @@ GenerateAST::Result GenerateAST::run(llvm::Function &func,
   // structurization
   llvm::ReversePostOrderTraversal<llvm::Function *> rpo(&func);
   rpo_walk.assign(rpo.begin(), rpo.end());
-  GetOrCreateReachingCond(rpo_walk[0]);
+  do {
```
surovic, line 619: Can you briefly describe in a comment why this fixpoint computation is needed?

lib/AST/GenerateAST.cpp
```diff
@@ -236,36 +236,68 @@ z3::expr GenerateAST::GetOrCreateEdgeCond(llvm::BasicBlock *from,
...
+  }};
+
+  auto Simplify{[this, &Prove](z3::expr expr) {
+    if (Prove(expr)) {
+      return z_ctx->bool_val(true);
+    }
+
+    z3::tactic aig(*z_ctx, "aig");
+    z3::tactic simplify(*z_ctx, "simplify");
+    z3::tactic ctx_solver_simplify(*z_ctx, "ctx-solver-simplify");
```
surovic, line 263: This is a pretty heavy duty simplifier :eyes: Have you tried `ctx-simplify`? Maybe mix `solve-eqs` in?

include/rellic/AST/GenerateAST.h
```diff
@@ -61,7 +62,8 @@ class GenerateAST : public llvm::AnalysisInfoMixin<GenerateAST> {
                                     llvm::ConstantInt *c);
 
   z3::expr GetOrCreateEdgeCond(llvm::BasicBlock *from, llvm::BasicBlock *to);
-  z3::expr GetOrCreateReachingCond(llvm::BasicBlock *block);
+  z3::expr GetReachingCond(llvm::BasicBlock *block);
+  void CreateReachingCond(llvm::BasicBlock *block);
 
   clang::Expr *ConvertExpr(z3::expr expr);
```
surovic, line 68: Have you considered using the existing code in `Z3ConvVisitor` directly?

surovic (changes requested):

include/rellic/AST/GenerateAST.h
frabert: `Z3ConvVisitor` converts each expression recursively, and simplifying the resulting expression causes a mess when trying to recover where the original components came from (see the Z3->C fiasco). I wanted a very simple conversion procedure, where I always have a 1:1 correspondence between Z3 variables and LLVM values (almost, `switch` conditions are slightly different)

surovic: Fair enough!

lib/AST/GenerateAST.cpp
frabert: `solve-eqs` seems to do the trick!

lib/AST/GenerateAST.cpp
frabert: Done!

lib/AST/GenerateAST.cpp
frabert: Done!

pgoodman (approved):

surovic (approved):

PR #267, opened 2022-05-23
Collect nested `if`s

frabert:


frabert:
Now the diff actually works as intended :)

pgoodman (approved):

PR #271, opened 2022-05-25
Speed up diff generation

frabert:


pgoodman:
Repeated conditions tested in `if`?

```
+            if ((int)val3 % 3 != 0U || !((int)val4 % 5 == 0U || (int)val3 % 3 != 0U)) {
```

frabert:
Simplifying a bit, that condition becomes `val3 % 3 != 0U || (val4 % 5 != 0U && val3 % 3 == 0U)`, so in this specific case I don't think there is any more elision to be applied

PR #277, opened 2022-06-26
Minimize suffixes when generating headers

frabert:


pgoodman:
I see:

```c
if ((int)val3 % 3 != 0U || !((int)val3 % 3 != 0U || (int)val4 % 5 == 0U)) {
```

As:

```
A || !(A || B)
```
```
A || (!A && !B) 
```
Which I believe is just:
```
A || !B
```
Due to the short-circuiting nature of `||`.

pgoodman (approved):

frabert:
The missed optimization opportunity is probably due to the new ZCS being kind of nearsighted I guess. I'm trying to figure out a way to spread Z3 info to more places.

re: shortcircuiting operators
All boolean operators should be already considered shortcircuiting during simplification, because those are translated straight to Z3 expressions. We can do that because all things that could produce side effects are pushed before the evaluation of conditions.

PR #280, opened 2022-07-19
Fix crashes from AMP challenges

frabert:
Should fix the crashes in Anvill-produced bitcode

cc @tetsuo-cpp


frabert:
Ahhh, so _now_ is when we start to see the ugliness of opaque pointers...

PR #282, opened 2022-07-21
Z3 all the things

frabert:
Converts Rellic to operate on Z3 expressions for as long as possible, only materializing to Clang AST before expression simplification.

Also keeps conditions in CNF without exponential blowup with a neat trick*

*it's still quadratic tho


frabert:
Also removes a bunch of options from the external API. I figured nobody was actually using them, and if someone actually needs the amount of control it provided, they may as well use the passes manually

frabert:
Unfortunately this still seems way too slow to be actually usable, especially on some Anvill-produced modules

surovic (changes requested):
Overall great work! May the Z3ConvVisitor rest in piece.

PS: I haven't gotten to `GenerateAST.(h|cpp)` or `IRToASTVisitor.(h|cpp)` yet, but could you add some short comments to the affected parts? Think "if there's a blank line, I might as well use it for a short comment describing what follows".

include/rellic/AST/Util.h
```diff
@@ -58,13 +59,33 @@ using IRToStmtMap = std::unordered_map<llvm::Value *, clang::Stmt *>;
...
+  Z3CondMap conds;
+
+  clang::Expr *marker_expr;
+
+  std::unordered_map<unsigned, BrEdge> z3_br_edges_inv;
+  std::map<BrEdge, unsigned> z3_br_edges;
+
+  std::unordered_map<llvm::SwitchInst *, unsigned> z3_sw_vars;
+  std::unordered_map<unsigned, llvm::SwitchInst *> z3_sw_vars_inv;
+  std::map<SwEdge, unsigned> z3_sw_edges;
```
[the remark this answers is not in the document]
frabert: `SwEdge` and `BBEdge` are `std::pair`s that don't have a hash specialization :/

include/rellic/Decompiler.h
```diff
@@ -20,35 +20,6 @@ namespace rellic {
 struct DecompilationOptions {
```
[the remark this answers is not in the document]
frabert: Nobody was using these options, and they become a chore to keep around as we shuffle around passes. In practice, if someone wants the full flexibility of choosing the pass order, they can do so by instantiating those themselves, like `rellic-xref` does. This API should be a more high-level "just do the thing".

lib/AST/CondBasedRefine.cpp
```diff
@@ -11,85 +11,91 @@
...
           new_else_body.push_back(else_b);
         }
-        new_if->setElse(ast.CreateCompoundStmt(new_else_body));
+
+        auto new_else{ast.CreateCompoundStmt(new_else_body)};
+        new_if->setElse(new_else);
       }
-    } else if (Prove(*z3_ctx, cond_a == !cond_b)) {
+
+      dec_ctx.conds[new_if] = dec_ctx.conds[if_a];
```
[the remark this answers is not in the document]
frabert: Hmm if we were to factor it out, we'd have to either put it in a `if(Prove(a == b) || Prove(a == !b))` which would force a preemptive evaluation of both, or put it uncoditionally before, which would create a lot of potentially temporary statements that never get put anywhere

lib/AST/CondBasedRefine.cpp
```diff
@@ -11,85 +11,91 @@
...
       body[i] = new_if;
       body.erase(std::next(body.begin(), i + 1));
       did_something = true;
     }
   }
-
   if (did_something) {
-    substitutions[compound] = ast.CreateCompoundStmt(body);
+    auto new_compound{ast.CreateCompoundStmt(body)};
```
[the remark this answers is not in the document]
frabert: Nope, reverted 😅

include/rellic/AST/Util.h
```diff
@@ -79,13 +100,16 @@ void CopyProvenance(TKey1 *from, TKey2 *to,
...
-z3::goal ApplyTactic(z3::context &ctx, const z3::tactic &tactic, z3::expr expr);
+z3::goal ApplyTactic(const z3::tactic &tactic, z3::expr expr);
+
+bool Prove(z3::expr expr);
 
-bool Prove(z3::context &ctx, z3::expr expr);
+z3::expr HeavySimplify(z3::expr expr);
+z3::expr_vector Clone(z3::expr_vector &vec);
 
+// Tries to keep each subformula sorted by its id so that they don't get
```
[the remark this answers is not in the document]
frabert: I went with `OrderById` for even more clarity, even if it's a bit ungrammatical maybe

lib/AST/NestedCondProp.cpp
```diff
@@ -12,106 +12,208 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 
+#include <unordered_map>
 #include <vector>
 
 #include "rellic/AST/ASTBuilder.h"
 #include "rellic/AST/Util.h"
 
+namespace std {
```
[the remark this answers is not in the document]
frabert: I went with just mapping the `id`s instead, they are supposed to be unique and we don't have to store the expressions anywhere

frabert:
There wasn't much change in `GenerateAST` or `IRToASTVisitor`, I mainly just moved the Z3->C conversion from one to the other because it now needs to happen later in the process.

lib/AST/CondBasedRefine.cpp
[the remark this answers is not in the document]
surovic: What about putting it after, behind a `if(did_something)`?

lib/AST/NestedCondProp.cpp
```diff
@@ -12,106 +12,208 @@
...
+      }
+      return;
+    }
+
+    values[expr.id()] = value;
+  }
+
+  // Simplify an expression `expr` using all the known values stored. Sets
+  // `found` to true is any simplification has been applied.
+  z3::expr ApplyAssumptions(z3::expr expr, bool& found) {
```
surovic, line 75: Minor nitpick, I'd rename `found` to `changed` to be a bit clearer.

lib/AST/NestedCondProp.cpp
```diff
@@ -12,106 +12,208 @@
...
 
 #include "rellic/AST/ASTBuilder.h"
 #include "rellic/AST/Util.h"
 
 namespace rellic {
-using ExprVec = std::vector<clang::Expr*>;
+// Stores a set of expression that have a known value, so that they can be
+// recognized as part of larger expressions and simplified.
+struct KnownExprs {
+  std::unordered_map<unsigned, bool> values;
```
surovic, line 25: What about using the hash specialisation trick here too? Or better yet, make an `ExprMap` and put it in `AST/Utils.cpp`.

lib/AST/CondBasedRefine.cpp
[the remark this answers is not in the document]
frabert: Done!

lib/AST/NestedCondProp.cpp
```diff
@@ -12,106 +12,208 @@
...
 
+#include <unordered_map>
 #include <vector>
 
 #include "rellic/AST/ASTBuilder.h"
 #include "rellic/AST/Util.h"
 
 namespace rellic {
-using ExprVec = std::vector<clang::Expr*>;
+// Stores a set of expression that have a known value, so that they can be
```
surovic, line 22: Just a thought, but do you think `z3::expr::substitute` could simplify this pass? I have doubts it would, but worth a question.

lib/AST/NestedCondProp.cpp
frabert: The issue with using the hash specialization trick is that `unordered_map` still needs to store the actual `z3::expr` objects in order to call the specializations on them, and at that point we're back to having to use an `expr_vector` I think

lib/AST/NestedCondProp.cpp
frabert: I tried to use it, and it didn't work for reasons I can't remember right now and should have probably documented 🤦 One problem I remember is that it didn't give an indication of whether it actually did anything, which meant you had to check for equality every time, which is slow as molasses

surovic: > it didn't give an indication of whether it actually did anything
True. Defintiely not worth it then .

lib/AST/NestedCondProp.cpp
surovic: Ah, fair.

include/rellic/AST/Util.h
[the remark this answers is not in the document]
surovic: Hmm, would it be possible to add [hash specialization](https://stackoverflow.com/a/20602159) for those? But don't bother if you think it's not worth it.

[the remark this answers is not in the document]
frabert: Can we just add a warning to investigate it if it causes performance issues down the line? Right now it works ok, and hashing code is ugly 😬

lib/AST/GenerateAST.cpp
```diff
@@ -136,78 +137,119 @@ static std::string GetName(llvm::Value *v) {
...
-  return z_exprs[z_br_edges[{inst, cond}]];
+  return dec_ctx.z3_br_edges[{inst, cond}];
 }
 
-z3::expr GenerateAST::GetOrCreateEdgeForSwitch(llvm::SwitchInst *inst,
+unsigned GenerateAST::GetOrCreateVarForSwitch(llvm::SwitchInst *inst) {
+  // To aide simplification, switch instructions actually produce numerical
+  // variables instead of boolean ones, but are always compared against a
+  // constant value.
+  if (dec_ctx.z3_sw_vars.find(inst) == dec_ctx.z3_sw_vars.end()) {
```
surovic, line 175: Let's do an early return here.

lib/AST/GenerateAST.cpp
```diff
@@ -136,78 +137,119 @@ static std::string GetName(llvm::Value *v) {
...
+unsigned GenerateAST::GetOrCreateEdgeForSwitch(llvm::SwitchInst *inst,
                                                llvm::ConstantInt *c) {
-  if (z_sw_edges.find({inst, c}) == z_sw_edges.end()) {
+  auto ToExpr = [&](unsigned idx) {
+    if (idx == poison_idx) {
+      return dec_ctx.z3_ctx.bool_val(false);
+    }
+    return dec_ctx.z3_exprs[idx];
+  };
+  if (dec_ctx.z3_sw_edges.find({inst, c}) == dec_ctx.z3_sw_edges.end()) {
```
surovic, line 195: Same as above. Early return.

lib/AST/GenerateAST.cpp
frabert: Would
```
  if (dec_ctx.z3_sw_vars.find(inst) == dec_ctx.z3_sw_vars.end()) {
    auto name{GetName(inst)};
    auto var{dec_ctx.z3_ctx.int_const(name.c_str())};
    dec_ctx.z3_sw_vars[inst] = dec_ctx.z3_exprs.size();
    dec_ctx.z3_exprs.push_back(var);
    dec_ctx.z3_sw_vars_inv[var.id()] = inst;
  }
  return dec_ctx.z3_sw_vars[inst];
```
be ok?

lib/AST/GenerateAST.cpp
```diff
@@ -136,78 +137,119 @@ static std::string GetName(llvm::Value *v) {
...
+    return dec_ctx.z3_sw_vars[inst];
+  } else {
+    return dec_ctx.z3_sw_vars[inst];
+  }
+}
+
+unsigned GenerateAST::GetOrCreateEdgeForSwitch(llvm::SwitchInst *inst,
                                                llvm::ConstantInt *c) {
-  if (z_sw_edges.find({inst, c}) == z_sw_edges.end()) {
+  auto ToExpr = [&](unsigned idx) {
```
surovic, line 189: `ToExpr` seems to be defined identically in a bunch of places. Let's factor it out into a private method.

include/rellic/AST/GenerateAST.h
```diff
@@ -25,28 +26,15 @@ class GenerateAST : public llvm::AnalysisInfoMixin<GenerateAST> {
   friend llvm::AnalysisInfoMixin<GenerateAST>;
   static llvm::AnalysisKey Key;
 
+  constexpr static unsigned poison_idx = std::numeric_limits<unsigned>::max();
```
surovic, line 29: What's the purpose of the `poison_idx`?

lib/AST/GenerateAST.cpp
surovic: I usually prefer it the other way around cause usually the big block of code is gonna contain long lines that benefit from one less indent. So, `if (dec_ctx.z3_sw_vars.find(inst) != dec_ctx.z3_sw_vars.end()) { return ...; } ...`

include/rellic/AST/GenerateAST.h
frabert: It's the index of an expression that has not been found. See e.g. `GetReachingCond`

lib/AST/IRToASTVisitor.cpp
```diff
@@ -61,9 +63,86 @@ class ExprGen : public llvm::InstVisitor<ExprGen, clang::Expr *> {
...
+
+    LOG(FATAL) << "Couldn't find switch case";
+  }
+
+  auto hash{expr.id()};
+  if (dec_ctx.z3_br_edges_inv.find(hash) != dec_ctx.z3_br_edges_inv.end()) {
+    auto edge{dec_ctx.z3_br_edges_inv[hash]};
+    CHECK(edge.second) << "Inverse map should only be populated for branches "
+                          "taken when condition is true";
+    return CreateOperandExpr(*(edge.first->op_end() - 3));
```
surovic, line 97: Does `*(edge.first->op_end() - 3)` return the branch condition? Definitely needs a comment or a more idiomatic way of getting the operand :eyes:

lib/AST/IRToASTVisitor.cpp
```diff
@@ -61,9 +63,86 @@ class ExprGen : public llvm::InstVisitor<ExprGen, clang::Expr *> {
   clang::Expr *visitUnaryOperator(llvm::UnaryOperator &inst);
 };
 
+clang::Expr *IRToASTVisitor::ConvertExpr(z3::expr expr) {
```
surovic, line 66: Can we add short comments on what cases each piece of code handles here?

lib/AST/IRToASTVisitor.cpp
```diff
@@ -61,9 +63,86 @@ class ExprGen : public llvm::InstVisitor<ExprGen, clang::Expr *> {
...
+                          "taken when condition is true";
+    return CreateOperandExpr(*(edge.first->op_end() - 3));
+  }
+
+  if (dec_ctx.z3_sw_vars_inv.find(hash) != dec_ctx.z3_sw_vars_inv.end()) {
+    auto inst{dec_ctx.z3_sw_vars_inv[hash]};
+    return CreateOperandExpr(inst->getOperandUse(0));
+  }
+
+  std::vector<clang::Expr *> args;
```
surovic, line 105: Is this piece of code really necessary? It seems like the places where it's used we can easily just use `ConvertExpr(expr.arg(i))`, or am I missing something?

frabert: Nope, taken out!

lib/AST/IRToASTVisitor.cpp
frabert: I couldn't find public API for what I needed, so I added a comment

lib/AST/IRToASTVisitor.cpp
frabert: Added some clarifications

frabert:
I think this is ready for merge now

surovic (approved):

PR #285, opened 2022-07-22
[WIP] Type inference on bitcode

frabert:


frabert:
I'm scrapping this for now. I think the method has some merit, but it needs more finesse and time to put into it, and at that point we might as well do it properly

PR #288, opened 2022-08-10
Limit amount of temporary variables

frabert:
Rellic generates temporary variables for every load in order to maintain an SSA-like structure. This also produces a lot of temporary variables which are annoying to read.

The observation is that a temporary is needed for a `load` of an `alloca` only if a `store` occurred between the use of said `load` and the `load` itself.

E.g.

```llvm
%alloca = alloca ...
store sth, %alloca
%load = load %alloca
%foo = add %load, ... ; %load doesn't need a temporary variable
```

```llvm
%alloca = alloca ...
store sth, %alloca
%load = load %alloca
store sthelse, %alloca
%foo = add %load, ... ; %load needs a temporary variable
```

### DIFF
--- a/lib/AST/CondBasedRefine.cpp
+++ b/lib/AST/CondBasedRefine.cpp
@@ -11,23 +11,11 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 
+#include <cstddef>
+
+#include "rellic/AST/Util.h"
+
 namespace rellic {
-
-namespace {
-
-using IfStmtVec = std::vector<clang::IfStmt *>;
-
-static IfStmtVec GetIfStmts(clang::CompoundStmt *compound) {
-  IfStmtVec result;
-  for (auto stmt : compound->body()) {
-    if (auto ifstmt = clang::dyn_cast<clang::IfStmt>(stmt)) {
-      result.push_back(ifstmt);
-    }
-  }
-  return result;
-}
-
-}  // namespace
 
 CondBasedRefine::CondBasedRefine(Provenance &provenance, clang::ASTUnit &unit)
     : TransformVisitor<CondBasedRefine>(provenance, unit),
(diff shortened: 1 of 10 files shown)


lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +24,49 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
-    std::vector<clang::Stmt *> new_body;
-    for (auto stmt : compound->body()) {
-      if (stmt) {
-        new_body.push_back(stmt);
+  std::vector<clang::Stmt *> body{compound->body_begin(), compound->body_end()};
+  std::vector<clang::IfStmt *> ifs;
+  z3::expr_vector conds{*z3_ctx};
+  bool done_something{false};
+  for (size_t i{0}; i < body.size(); ++i) {
+    if (auto if_stmt = clang::dyn_cast<clang::IfStmt>(body[i])) {
```
surovic, line 33: Try to do an early `continue` here.

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +24,49 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+      ifs.push_back(if_stmt);
+      auto cond{GetZ3Cond(if_stmt)};
+      if (!if_stmt->getElse() && Prove(*z3_ctx, !(cond && z3::mk_or(conds)))) {
```
surovic, line 36: I would preserve the `IsUnreachable` lambda here to make it a bit more clear what we're trying to prove wrt. to the reachind conditions.

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +24,49 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+        conds.push_back(GetZ3Cond(if_stmt));
+
+        if (Prove(*z3_ctx, z3::mk_or(conds)) && ifs.size() > 2) {
```
surovic, line 39: What are we proving here? In the previous version we tried to determine if the reaching conditions form a tautology, which signifies that we have everything we need to create an `if-else` cascade. AFAICT this checks whether the `or` of the reaching conditions is satisfiable, not a tautology.

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +24,49 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+          auto last_if{ifs[0]};
```
surovic, line 40: There should be a comment here saying what this piece of code does. I assume it's the creation of the `if-else` cascade, right?

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +24,49 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+            }
+            if (stmt == ifs.back()) {
+              last_if->setElse(stmt->getThen());
+            } else {
+              last_if->setElse(stmt);
+              last_if = stmt;
+            }
+          }
+
+          size_t start_delete{i - (ifs.size() - 2)};
```
surovic, line 53: Same as above. This one removes the `if` statements that were put in the cascade, right?

surovic (changes requested):
The new condition-based refinement pass is fine IMO. Better and simpler than the previous version, but might miss some structurization opportunities compared to the version in No More Goto's or the [Dream](https://github.com/CodeIntelligenceTesting/dream/blob/ef467aa516ae5dc5a466cad11887f517dd302c08/dream/ControlFlowTree.py#L661) decompiler, which does clustering across the whole compound statement, but also is aware of reachability and probably side-effects too in some way.

The reach-based refine seems conceptually similar to me, but might have correctness issues that I mentioned in the comments in the code

lib/AST/ReachBasedRefine.cpp
frabert: `Prove(P)` returns `true` if `!P` is proven unsatisfiable, i.e. if `P` is proven true under all circumstances

lib/BC/Util.cpp
```diff
@@ -437,4 +437,61 @@ void ConvertArrayArguments(llvm::Module &m) {
...
+    std::vector<llvm::Instruction *> work_list;
+    auto inst{llvm::cast<llvm::Instruction>(user)};
+    work_list.push_back(inst);
+    while (!work_list.empty()) {
+      inst = work_list.back();
+      work_list.pop_back();
+      auto bb{inst->getParent()};
+      visited_blocks.insert(bb);
+
+      while (inst) {
```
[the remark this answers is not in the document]
frabert: Hrmm I can't see an easy way of doing that, because I also need to navigate through the predecessor blocks

frabert:
For some reason Z3 is not able to recognize the cascading conditions in `switch_loop` without the context aware simplifier, but that is too slow to use in that place.

lib/AST/CondBasedRefine.cpp
```diff
@@ -41,85 +29,65 @@ z3::expr CondBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+        new_if->setElse(ast.CreateCompoundStmt(new_else_body));
       }
-      // Add the else branch
-      sub->setElse(ast.CreateCompoundStmt(elses));
-    }
-    // Replace `lhs` with the new `sub`
-    substitutions[lhs] = sub;
-  }
-}
+      body[i] = new_if;
```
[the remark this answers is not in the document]
frabert: Done!

lib/AST/ReachBasedRefine.cpp
frabert: I have factored out the conditions to variables, do you think it's explicit enough now?

lib/AST/ReachBasedRefine.cpp
frabert: Refactored the loop

lib/AST/ReachBasedRefine.cpp
frabert: Added comments

lib/AST/ReachBasedRefine.cpp
frabert: Ditto

lib/BC/Util.cpp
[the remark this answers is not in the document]
surovic: What I'm a bit worried about is that testing `while(inst)` that's moved by `inst = inst->getPrevNode()` is gonna run out of bounds. Hmm, what about instead of putting just the terminator into the worklist, you'd put in the whole predecessor BB in reverse via `work_list.push_back(pred.rbegin(), pred.rend())` ? And initialize via `inst->getReverseIterator()` and `inst->getParent()->rend()`.

lib/AST/CondBasedRefine.cpp
```diff
@@ -41,85 +29,67 @@ z3::expr CondBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+
+    auto else_a{if_a->getElse()};
+    auto else_b{if_b->getElse()};
+
+    auto cond_a{GetZ3Cond(if_a)};
+    auto cond_b{GetZ3Cond(if_b)};
+
+    clang::IfStmt *new_if{};
+    if (Prove(*z3_ctx, cond_a == cond_b)) {
+      std::vector<clang::Stmt *> new_then_body{then_a, then_b};
```
surovic, line 54: I think you can factor out this as well. I think everything besides that is fine.

lib/BC/Util.cpp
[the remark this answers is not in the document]
surovic: I know, but I'd still prefer using iterators, ranges and all that. I'm not gonna insist though.

lib/AST/ReachBasedRefine.cpp
surovic: Yup, this is great! Thanks!

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +26,116 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+bool ReachBasedRefine::VisitCompoundStmt(clang::CompoundStmt *compound) {
+  std::vector<clang::Stmt *> body{compound->body_begin(), compound->body_end()};
+  std::vector<clang::IfStmt *> ifs;
+  z3::expr_vector conds{*z3_ctx};
+
+  bool done_something{false};
+  for (size_t i{0}; i < body.size(); ++i) {
+    auto if_stmt{clang::dyn_cast<clang::IfStmt>(body[i])};
+    if (!if_stmt) {
+      ifs.clear();
```
surovic, line 38: Maybe factor this out into a helper lambda, since it repeats a bunch of times.

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +26,116 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+    // Do the collected statements cover all possibilities?
+    auto is_complete{Prove(*z3_ctx, z3::mk_or(conds))};
+
+    if (ifs.size() <= 2 || !is_complete) {
+      // We need to collect more statements
+      continue;
+    }
+
+    /*
+    `body` will look like this at this point:
```
surovic, line 73: This is my subjective preference so take it with a pinch of salt, but I usually prefer comments to be short / dense so that it doesn't break the flow of the code and a relevant scope or function is on one editor screen. That being said, the comment is excellent so it'd be a shame to change it.

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +26,116 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+        last_if->setElse(stmt->getThen());
+      } else {
+        last_if->setElse(stmt);
+        last_if = stmt;
       }
     }
-    substitutions[compound] = ast.CreateCompoundStmt(new_body);
+
+    /*
+    `body` will look like this at this point:
```
surovic, line 108: Same as above.

lib/AST/ReachBasedRefine.cpp
```diff
@@ -40,83 +26,116 @@ z3::expr ReachBasedRefine::GetZ3Cond(clang::IfStmt *ifstmt) {
...
+      i - n    : ...
+      i - n + 1: if(cond_1) { } else if(cond_2) { } else if ...
+      ...
+    */
+    size_t start_delete{i - (ifs.size() - 2)};
+    size_t end_delete{i};
+    body.erase(body.erase(std::next(body.begin(), start_delete),
+                          std::next(body.begin(), end_delete)));
+    done_something = true;
+    break;
```
surovic, line 132: I'd replace this `break` with adding `... && !done_something` in the loop's continuation cond. But it's only a nitpick.

lib/AST/ReachBasedRefine.cpp
frabert: Yeah I figured the best way to show what the code was doing was a little "picture", since indices are confusing otherwise

surovic:
LGTM. Can I ask you to make an issue for the CBR investigation I mentioned above? So that it's not forgotten.

surovic (approved):

lib/AST/CondBasedRefine.cpp
frabert: You mean `new_then_body`?

PR #303, opened 2022-09-20
Bring `ast_unit`, `ast_ctx`, `ast` into `dec_ctx`

frabert:


surovic (approved):
Just a small nitpick. In cases like `IRToASTVisitor` and I think `GenerateAST`, where there are functions that used `ast` a lot, it's probably a good idea to do something like `auto ast = dec_ctx.ast;` rather than doing a plain search and replace and generating a huge diff and also longer expressions :eyes:

frabert:
Done!

PR #307, opened 2022-10-17
Add `dec2hex` tool

frabert:


lib/Dec2Hex.cpp
```diff
@@ -0,0 +1,98 @@
...
+      lit->getValue().toString(
+          str, /*radix=*/16, /*isSigned=*/lit->getType()->isSignedIntegerType(),
+          /*formatAsCLiteral=*/true);
+      std::string res;
+      llvm::raw_string_ostream OS(res);
+      OS << str;
+      switch (lit->getType()->castAs<BuiltinType>()->getKind()) {
+        default:
+          llvm_unreachable("Unexpected type for integer literal!");
+        case BuiltinType::Char_S:
```
surovic, line 49: Suffixes containing `i` like `Ui8` are and MSVC extension iirc, but I think it's okay to leave it here?

frabert: This specific part is lifted straight from Clang, but I guess we could tune it to our liking if need be

PR #308, opened 2022-10-19
Migrate to LLVM15

frabert:


Ninja3047:
diff is failing to build due to llvm version mismatch, should get resolved after this gets merged

tools/xref/StmtPrinter.cpp
```diff
@@ -185,11 +186,72 @@ class StmtPrinter : public StmtVisitor<StmtPrinter> {
 /// with no newline after the }.
 void StmtPrinter::PrintRawCompoundStmt(CompoundStmt *Node) {
   OS << "{" << NL;
+  PrintFPPragmas(Node);
   for (auto *I : Node->body()) PrintStmt(I);
 
   Indent() << "}";
 }
 
+void StmtPrinter::PrintFPPragmas(CompoundStmt *S) {
```
surovic, line 195: I assume this is a straight copy-paste from clang, right?

tools/xref/StmtPrinter.cpp
```diff
@@ -751,13 +813,25 @@ void StmtPrinter::VisitOMPParallelMasterDirective(
   PrintOMPExecutableDirective(Node);
 }
 
+void StmtPrinter::VisitOMPParallelMaskedDirective(
```
surovic, line 816: Do we need the OpenMP pragma printers for anything?

surovic (approved):
Just to be sure, the changes in various printers are straight copy-pastes from clang, right? Otherwise LGTM.

frabert:
All copy-pastes. I want to get rid of the custom printer when we update to LLVM16, but that might require using PASTA, which requires a patched Clang

tools/xref/StmtPrinter.cpp
frabert: Just for making Clang happy and making tracking changes across versions easier

PR #309, opened 2022-10-24
Migrate CI to Github Large Runner

Ninja3047:


Ninja3047:
Should be passing most of checks, i think a few need some more work

frabert:
I think the AWS-related secrets can be removed

Ninja3047:
Ah yeah, wasn't sure if we still wanted to upload the test artifacts to s3
CC @artemdinaburg

Ninja3047:
AWS related secrets are DO spaces credentials that we upload artifacts to through their S3 compatible API

frabert (approved):

PR #313, opened 2022-11-28
Move to explicitly setting CMake toolchain file

tetsuo-cpp:


.github/workflows/diff_tests.yml
```diff
@@ -62,13 +62,13 @@ jobs:
       - name: Build old rellic
         shell: bash
         run: |
-          cmake -G Ninja -S old -B rellic-build-old -DVCPKG_ROOT=$GITHUB_WORKSPACE/vcpkg_${{ matrix.image.name }}-${{ matrix.image.tag }}_llvm-${{ matrix.llvm }}_amd64
+          cmake -G Ninja -S old -B rellic-build-old -DCMAKE_TOOLCHAIN_FILE=$GITHUB_WORKSPACE/vcpkg_${{ matrix.image.name }}-${{ matrix.image.tag }}_llvm-${{ matrix.llvm }}_amd64/scripts/buildsystems/vcpkg.cmake
```
tetsuo-cpp, line 65: The test failure is caused by this line because `master` requires `VCPKG_ROOT`. However, once this gets merged it will begin working so I think we can just leave this as is.

frabert: That's ok, we have a similar issue regarding LLVM14 vs 15

CMakePresets.json
```diff
@@ -60,7 +60,7 @@
           "generator": "Ninja",
           "cacheVariables": {
               "VCPKG_TARGET_TRIPLET": "$env{VCPKG_TARGET_TRIPLET}",
-              "VCPKG_ROOT": "$env{VCPKG_ROOT}",
+              "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
```
tetsuo-cpp, line 63: Same deal here as with `remill` and `anvill`. If we want the presets to use the Clang that comes bundled with the LLVM install that we're building against, the presets need `VCPKG_ROOT`.

scripts/build.sh
```diff
@@ -19,7 +19,7 @@ CURR_DIR=$( pwd )
 BUILD_DIR="${CURR_DIR}/rellic-build"
 INSTALL_DIR=/usr/local
 LLVM_VERSION=llvm-14
-CXX_COMMON_VERSION=v0.2.10
+CXX_COMMON_VERSION=v0.2.16
```
tetsuo-cpp, line 22: This isn't strictly related, but I just bumped the version to get things building on my M1.

CMakePresets.json
frabert: I'm not sure I follow why the cache variable would be needed in addition to the env var?

tetsuo-cpp: A few lines below, the preset sets the C and C++ compilers to the ones in the LLVM install (so something like `${VCPKG_ROOT}/installed/$env{VCPKG_TARGET_TRIPLET}/tools/llvm/clang`). If we remove `VCPKG_ROOT` as a cache variable and only have `CMAKE_TOOLCHAIN_FILE`, I'm not sure how we can do that.

frabert: `"CMAKE_C_COMPILER": "$env{VCPKG_ROOT}/installed/$env{VCPKG_TARGET_TRIPLET}/tools/llvm/clang"` uses `$env{VCPKG_ROOT}`, not the cache var, or am I missing something?

tetsuo-cpp: Are you suggesting that the `CMAKE_TOOLCHAIN_FILE` cache variable could be removed because it isn't used to find the compilers?

frabert: `CMAKE_TOOLCHAIN_FILE` is used by Rellic to find the packages so it can't be removed. What confuses me is

> If we want the presets to use the Clang that comes bundled with the LLVM install that we're building against, the presets need `VCPKG_ROOT`.

I guess I don't understand which `VCPKG_ROOT` you're referring to, the env var or cache var?

tetsuo-cpp: > I guess I don't understand which `VCPKG_ROOT` you're referring to, the env var or cache var?

I'm referring to the env var. I've removed all references of `VCPKG_ROOT`, but this is one spot where I'm not sure that I can.

frabert: Ohhh sorry then 🤣 Keeping the `VCPKG_ROOT` is fine to me, the user needs to be able to specify where to find vcpkg in any case

pgoodman: I am confused, but probably out of ignorance because I don't know anything about these presets files. Why is `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` needed in the presets?

tetsuo-cpp: I don't think it's needed, it's just convenient. The presets are a handful of commonly used configurations and one way that they try to be helpful is by setting `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` to the Clang versions that belong to the LLVM install you're building against.

This gives you a relatively recent compiler which is probably helpful on systems where the default `cc` and `c++` are outdated. On the other hand, I think this is counterproductive for debug builds because you probably don't want to be building with a debug build of Clang.

I'd be happy to just get rid of this unless anyone else is opposed to it.

frabert: If removing that doesn't break anything else, I'm all for it

pgoodman: Ditto.

frabert (approved):

PR #324, opened 2023-05-16
Upgrade to LLVM16

frabert:


scripts/test-headergen.py
```diff
@@ -94,12 +97,14 @@ class TestRoundtrip(unittest.TestCase):
     parser.add_argument("tests", help="path to test directory")
     parser.add_argument("clang", help="path to clang")
     parser.add_argument("-t", "--timeout", help="set timeout in seconds", type=int)
+    parser.add_argument(
```
surovic, line 100: I'd probably add a more generic "additional flags" parameter here. Adding mac OS specific flags feels wrong. That being said, it's more of a preference thing.

Issue #325, opened 2023-05-18
Rellic produces semantically non-equivalent code


vmihalko:
##### Used version:
0.2.3. (current)

##### Reproducer:
```c
#include <stdio.h>
int main(int argc, const char** argv) {
	int i = argc;
	if ( i == 1 ) { // no args
		putchar(4 + '0');
		i += 1;
		goto n5;	
	}
	if ( i > 1 ) { // args
		putchar( 5 + '0');
		goto n7;
	}
n5:
	putchar( 6 + '0');
n7:
	putchar( 7 + '0');
}

```
##### Steps to reproduce:
```bash
clang error.c -o error
clang error.c -S -emit-llvm -o error.ll
rellic-decomp --input error.ll --output de-error.c
clang de-error.c -o de-error
```
##### Expected behavior:
Executing `error` results in: "467"
##### Current behavior:
But executing `de-error` results in: "47"

##### Rellics output:
```c
int main(int arg0, char **arg1);
unsigned int putchar(unsigned int arg0);
int main(int arg0, char **arg1) {
    unsigned int var0;
    unsigned int var1;
    char **var2;
    unsigned int var3;
    var0 = 0U;
    var1 = arg0;
    var2 = arg1;
    var3 = var1;
    if (var3 != 1U && (int)var3 > 1) {
        putchar(53U);
    }
    if (var3 == 1U) {
        putchar(52U);
        var3 = var3 + 1U;
    }
    if (var3 == 1U || (int)var3 <= 1) {
        putchar(54U);
    }
    putchar(55U);
    return var0;
}

```

frabert:
Can reproduce 👍

PR #326, opened 2023-05-19
Fix issue #325

frabert:


frabert:
@surovic Thoughts on this? The fix works (until I figure out why the test fails on Linux O1), but the resulting decompilate is very verbose. At the same time, it's fixable by letting mem2reg remove a bunch of allocas and loads...

frabert:
The issue is that just tracking the number of uses of an instruction doesn't work.

```llvm
define i32 @main(i32 noundef %0, ptr noundef %1) #0 {
  %3 = alloca i32, align 4
  %4 = alloca i32, align 4
  %5 = alloca ptr, align 8
  %6 = alloca i32, align 4
  store i32 0, ptr %3, align 4
  store i32 %0, ptr %4, align 4
  store ptr %1, ptr %5, align 8
  %7 = load i32, ptr %4, align 4
  store i32 %7, ptr %6, align 4
  %8 = load i32, ptr %6, align 4
  %9 = icmp eq i32 %8, 1
  br i1 %9, label %10, label %14

10:                                               ; preds = %2
  %11 = call i32 @putchar(i32 noundef 52)
  %12 = load i32, ptr %6, align 4
  %13 = add nsw i32 %12, 1
  store i32 %13, ptr %6, align 4
  br label %20

14:                                               ; preds = %2
  %15 = load i32, ptr %6, align 4
  %16 = icmp sgt i32 %15, 1
  br i1 %16, label %17, label %19

17:                                               ; preds = %14
  %18 = call i32 @putchar(i32 noundef 53)
  br label %22

19:                                               ; preds = %14
  br label %20

20:                                               ; preds = %19, %10
  %21 = call i32 @putchar(i32 noundef 54)
  br label %22

22:                                               ; preds = %20, %17
  %23 = call i32 @putchar(i32 noundef 55)
  %24 = load i32, ptr %3, align 4
  ret i32 %24
}
```

```c
int main(int arg0, char **arg1) {
    unsigned int var0;
    unsigned int var1;
    void *var2;
    unsigned int var3;
    var0 = 0U;
    var1 = arg0;
    var2 = arg1;
    var3 = var1;
    if (var3 != 1U && (int)var3 > 1) {
        putchar(0x35U);
    }
    if (var3 == 1U) {
        putchar(0x34U);
        var3 = var3 + 1U;
    }
    if (var3 == 1U || (int)var3 <= 1) {
        putchar(0x36U);
    }
    putchar(0x37U);
    return var0;
}
```

In this case, the bug would have been fixed if Rellic had created a storage variable for `%9`, but that didn't happen because it's only used once in the LLVM bitcode, yet its expression (`var3 == 1U`) is used multiple times during generation. Maybe we can track the number of uses of each instruction while generating the code.

pgoodman:
The code checks `hasNUsesOrMore(1)` on calls and loads. This surprises me; why are those conditional on at least one use at all? Is there some logic somewhere that is able to tell us that these instructions have no side-effects (e.g. the call being pure, or the the load being on an alloca or constant global)?

It seems like the underlying problem that the `hasNUsesOrMore` is trying to solve is to help fake the maintenance of the existing partial order among calls, loads, and stores induced by the control-flow graph (ignoring back-edges).

frabert:
In a sense, that is true. The issue is that Rellic's codegen is not order-preserving, so we need to "snapshot" the value of instructions that have potential side-effects at the moment they are evaluated.

Re: logic that discovers if instructions are free of side effects

I'm not aware of such mechanism being already available in LLVM, but cooking up something ad-hoc for our specific use case seems feasible

Re: conditionals

For calls, it's so we cache the return value of function calls in case they are used elsewhere, while still keeping calls that do not return stuff on their own. For loads, I'd say it's a bug introduced by yours truly :)

pgoodman:
Is there a way to encode the partial order relation as Z3 expressions? That is, could `load`s be somehow turned into uninterpreted function calls that return a tuple, where one element is the value returned, and the other is a "memory token" (think Remill's memory pointer). Then you could even fake a `phi` node in z3 if you needed to merge multiple tokens.

frabert:
I'm not sure what that would look like. Could you make a quick example?

pgoodman:
Define a new LLVM function with prototype roughly

```
ptr __ptr(ptr, ptr)
```

Transform loads as follows:
```llvm
%x = load %y
```
Into:
```llvm
%MEMORY = alloca ptr
...
%y_memory = load ptr %MEMORY
%y1 = call __ptr(%y, %y_memory)
store %y1, %MEMORY
%x = load %y1
```

Or possibly define `__ptr` as: `ptr __ptr(ptr, ...)` and "mix in" the loaded values themselves, to really enforce ordering:

```llvm
%y_memory = load ptr %MEMORY
%y1 = call __ptr(%y, %y_memory)
%x = load %y1
%y2 = call __ptr(%y1, %x)
store %y2, %MEMORY
```

Then run mem2reg on this to eliminate the `%MEMORY` `alloca`, then add custom lift logic to ignore `__ptr`, or always lift it as an `NullStmt` or something like that. The "something like that" is load bearing here; I don't know how precisely you represent this as ASTs so that is both "there" but "not there."

On the assumption that the above is viable, I predict two kinds of contention with this approach:
 1) This changes the bitcode.
 2) This introduces an ad hoc, informally specified dialect into the bitcode.

The latter is a stronger argument against this approach, but I think that argument fails at the point that the whole process Rellic goes through is a bunch of unspecified dialects of Clang ASTs. I recall that intermediate stages of Rellic's procedures produce code that is not valid, and requires some simplification to eliminate issues. The former is a means to an end, like running mem2reg before Rellic, which practically all uses cases do anyway, so I don't think it's really a reasonable blocker.

frabert:
Honestly I'm not convinced this approach is going to be particularly better or easier to implement than tracking the usage of loads more accurately.

This seems to be introducing complexity to the "preprocessing stage" that can potentially make debugging additional issues more difficuly (more levels of indirection), and it seems to be solving the issue by basically turning loads into calls (which we currently handle correctly, as far as we know at least) but would introduce lots of auxiliary variables as well

PR #328, opened 2023-07-06
Replace undef constant floats with 0

2over12:
NaN would also be kinda viable


frabert:
The patch looks good to me, I'm just confused as to why the CI is not running 🙃

frabert (approved):

PR #330, opened 2023-09-26
llvm 17 fixes

2over12:


frabert:
Does this compile with LLVM16 still? If not, I think the CI needs changing as well

frabert:
Hrmm workflows don't seem to be working...

PR #331, opened 2023-09-28
Make built docker image runnable on Apple Silicon

konchunas:
Without this line the image built just fine, but when trying to run it failed with `qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory` error. I believe this parameter forces it to be build for the selected architecture so it is runnable under other archs.
Pretty sure this wont break building docker in Linux, but I haven't tested it.


frabert (approved):

frabert:
Thank you!

PR #340, opened 2025-05-16
Support for llvm 20

kumarak:
The PR changes support to build with llvm 20 and cmake changes to move away from cxx-common dependency

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,8 @@ endif(NOT DEFINED WIN32)
 # libraries
 #
 
+# gflags needs this to create the gflags::gflags target that glog expects
+set(GFLAGS_USE_TARGET_NAMESPACE ON)
 find_package(gflags CONFIG REQUIRED)
 find_package(glog CONFIG REQUIRED)
 find_package(Z3 4.8 CONFIG REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,9 +50,15 @@ find_package(gflags CONFIG REQUIRED)
 find_package(glog CONFIG REQUIRED)
 find_package(Z3 4.8 CONFIG REQUIRED)
 find_package(doctest CONFIG REQUIRED)
-find_package(LLVM CONFIG REQUIRED)
+find_package(LLVM 20 CONFIG REQUIRED)
+message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 llvm_map_components_to_libnames(llvm_libs support core irreader bitreader bitwriter)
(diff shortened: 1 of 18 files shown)


kumarak:
Moving to draft since I need to fix CI as well.

CMakeLists.txt
```diff
@@ -50,9 +50,15 @@ find_package(gflags CONFIG REQUIRED)
 find_package(glog CONFIG REQUIRED)
 find_package(Z3 4.8 CONFIG REQUIRED)
 find_package(doctest CONFIG REQUIRED)
-find_package(LLVM CONFIG REQUIRED)
+find_package(LLVM 20 CONFIG REQUIRED)
+message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 llvm_map_components_to_libnames(llvm_libs support core irreader bitreader bitwriter)
-find_package(Clang CONFIG REQUIRED)
+
+find_package(MLIR 20 CONFIG REQUIRED)
```
frabert, line 59: Does Rellic need MLIR?

tools/CMakeLists.txt
```diff
@@ -57,14 +57,22 @@ add_executable(${RELLIC_XREF}
...
 find_path(CPP_HTTPLIB_INCLUDE_DIRS "httplib.h")
 target_include_directories(${RELLIC_XREF} PRIVATE ${CPP_HTTPLIB_INCLUDE_DIRS})
 
 target_link_libraries(${RELLIC_XREF}
   PRIVATE
     "${PROJECT_NAME}_cxx_settings"
     "${PROJECT_NAME}"
-    gflags::gflags
+    gflags
 )
```
ekilmer, line 76: IMO this should be gated behind an option like
```
option(RELLIC_FIND_ALL_EXTERNAL_DEPENDENCIES "Use find_package for ALL dependencies" OFF)
```
in `cmake/options.cmake`. At least let the user choose to provide their own cpp-httplib, even if it's untested by default. Most users won't care, which is fine, and it'll do the right thing because it's `OFF` by default.

Also, we should use the target `httplib::httplib` instead of `CPP_HTTPLIB_INCLUDE_DIRS`.

tools/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ target_link_libraries(${RELLIC_DECOMP}
   PRIVATE
     "${PROJECT_NAME}_cxx_settings"
     "${PROJECT_NAME}"
-    gflags::gflags
+    gflags
```
ekilmer: Why is this changing? gflags provides official support for `gflags::gflags`

ekilmer (changes requested):

PR #341, opened 2025-10-29
Comment out AnghaBench CI Cron Jobs workflow

artemdinaburg:
Comment out angha ci; it has been failing for a while and I do not have time to fix & investigate.


frabert:
Can we delete the file altogether? 🤔

artemdinaburg:
> Can we delete the file altogether? 🤔

I think this would indeed be cleaner, originally I thought to comment it in case someone fixed the test but its likely to just confuse more than help

frabert (approved):

